PR #22, opened 2017-10-02
Init Pass on  MaxSrcResolution for Issue 8

bscott:
Working PR to discuss the defaults for Resolution vs using Dimension.

### DIFF
--- a/config.go
+++ b/config.go
@@ -69,10 +69,10 @@ type config struct {
 	MaxClients      int
 	TTL             int
 
-	MaxSrcDimension int
-
-	Quality         int
-	GZipCompression int
+	MaxSrcDimension  int
+	MaxSrcResolution int
+	Quality          int
+	GZipCompression  int
 
 	Key  []byte
 	Salt []byte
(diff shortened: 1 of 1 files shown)


config.go
```diff
@@ -69,10 +69,10 @@ type config struct {
 	MaxClients      int
 	TTL             int
 
-	MaxSrcDimension int
-
-	Quality         int
-	GZipCompression int
+	MaxSrcDimension  int
+	MaxSrcResolution int
+	Quality          int
```
DarthSim: Don't remove separating empty lines, please.

config.go
```diff
@@ -111,7 +112,7 @@ func init() {
 	intEnvConfig(&conf.TTL, "IMGPROXY_TTL")
 
 	intEnvConfig(&conf.MaxSrcDimension, "IMGPROXY_MAX_SRC_DIMENSION")
-
+	intEnvConfig(&conf.MaxSrcResolution, "IMGPROXY_MAX_SRC_RESOLUTION")
```
DarthSim: Don't remove separating empty lines, please.

config.go
```diff
@@ -81,15 +81,16 @@ type config struct {
...
-	Quality:         80,
-	GZipCompression: 5,
+	Bind:             ":8080",
+	ReadTimeout:      10,
+	WriteTimeout:     10,
+	DownloadTimeout:  5,
+	Concurrency:      runtime.NumCPU() * 2,
+	TTL:              3600,
+	MaxSrcDimension:  4096,
+	MaxSrcResolution: 65,
```
DarthSim: 65 megapixels is quite hight for default IMHO. Let's set it to 15.

DarthSim:
And thanks for your contribution! :)

homm:
> `MaxSrcDimension` doesn't make much sense

I don't know VIPS internals but for some libraries, this does sense, because too tall or too wide images (for example 1x15000000) can have some overhead in processing and this could be used as a vulnerability.

I suggest setting the max resolution to some suitable value, like 15M and set the max dimension to a high, but realistic value like 16K or 32K.

config.go
homm: It used to be MaxSrcDimension* MaxSrcDimension, i.e. 16 megapixels.

DarthSim:
Closed with https://github.com/DarthSim/imgproxy/commit/2940478ec9ef298129a6e6df1c58e688cd03c478

PR #30, opened 2017-10-25
Add IMGPROXY_TTL env to readme

Mikulas:
1. Users should be made aware the endpoint allows caching
2. Depending on context users may want to increase the TTL to a very large number (for example when serving as an Origin for CloudFront Distribution or other CDN).


DarthSim:
Thanks again :)

Robx64:
Hello guys,
Where can I find any information about imgproxy requirements in regards to the hardware spec if it has any special requirements at all ?

DarthSim:
@Robx64 No it doesn't. imgproxy works perfectly even on free Heroku instance.

Robx64:
Great, thank you DS ;P

PR #31, opened 2017-10-25
Add /health endpoint

Mikulas:
Returns 200 OK after server starts.

This greatly simplifies container deployment. For example,
Kubernetes can be configured to not route traffic to imgproxy Pods
that did not pass Readiness check against /health. (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request)

Another option is returning generic about/help page for `/`. I figured an explicit `/health` endpoint is less intrusive and more readable in Readiness checks.


DarthSim:
Thanks for the contribution!

PR #32, opened 2017-10-27
Add swift signing example

Mikulas:
HMAC is slightly complex to use as it was not yet ported to pure Swift.

Can be complied and run as
```bash
echo '#import <CommonCrypto/CommonCrypto.h>' > bridge.h
swiftc examples/signature.swift -import-objc-header bridge.h
./signature
```


DarthSim:
Thanks for the contribution!

PR #33, opened 2017-10-30
correct Cache-contol header missusage.

bnprss:
Cache-control header value is currently wrongly set.
For a TTL set to 3600s, the raw header value is set to "Cache-Control: Cache-Control: max-age=3600".
The correction set the value to "Cache-Control: max-age=3600, public" to nicely fit most CDNs requirements.


DarthSim:
My bad. Thanks for the catch!

PR #34, opened 2017-11-13
Add Elixir example

princemaple:
@DarthSim Thanks for the great tool.


DarthSim:
Thanks for the contribution!

PR #41, opened 2017-12-29
Add option for reading images from local filesystem

Vasfed:
By default go's net/http does not support `file://` urls for security reasons. But for some use cases it may be proficient to be able to read source images from local filesystem.

This PR adds `IMGPROXY_LOCAL_FILESYSTEM_ROOT` config variable for `local://` urls, disabled by default.


tangentlin:
Nice, this is a great solution for feature request #42

DarthSim:
Hi!

I totally love the PR! It adds the new feature without complicating of the code and without changing workflow in any form.

PR #43, opened 2018-01-11
Fix cmyk profile path

Vasfed:
This fixes a typo in cmyk profile loading that causes
```
| 500 | VipsIcc: File 'cmyk.icm' not found
icc_import: unable to open profile "cmyk.icm"
icc_import: no input profile
```


DarthSim:
Omg, thank you

PR #47, opened 2018-01-31
Add  support of ETag caching

w32blaster:
# Intro
I am using Imgproxy to serve local image files and all the requests are cached by CDN (KeyCDN). During debugging I noticed, that images are always downloaded from my CDN with status 200. But as long as I change my images relatively rarely, I heavily rely on browser's local cache

# Solution
I decided to add support of ETag header that is the best solution to provide the caching functionality in this scenario. Furthermore, many CDN respects these headers. I implemented very simple logic that checks the sha1 sum of a current image and in case if a user has a valid locally cached copy of the same image, then server returns 304 ("Not Modified") instead of sending the whole image again and again. 

This functionality is disabled by default. To enable it, you need to provide the `IMGPROXY_CACHE_FILES="true"` environment variable to activate this setting.

The caching works both for remote and local files. Tested locally and using KeyCDN service.

selection_102

P.S. Thank you for this project! :+1:


README.md
```diff
@@ -221,7 +223,7 @@ You can find helpful code snippets in the `examples` folder.
 imgproxy can process files from your local filesystem. To use this feature do the following:
 
 1. Set `IMGPROXY_LOCAL_FILESYSTEM_ROOT` to your images directory path.
-2. Use `local:///path/to/image.jpg` as the source image url.
+2. Use `local:///path/to/image.jpg` as the source image url. The path should be relative to the filesystem root value
```
w32blaster: Sorry, I added this sentence, because I wasted half an hour trying to understand why it doesn't work for me. Turns out I set absolute path which is wrong.

DarthSim: This shouldn't be a part of this PR

config.go
```diff
@@ -30,6 +30,18 @@ func strEnvConfig(s *string, name string) {
...
+	*b = false
+	if env := os.Getenv(name); len(env) > 0 {
+		bVal, err := strconv.ParseBool(env)
+		if err != nil {
+			*b = false
+			return
+		}
+		*b = bVal
+	}
+}
```
DarthSim: Redundant code for such simple action. This should be enough:

```go
if env, err := os.Getenv(name); err == nil {
  *b = env
}
```

It could be implemented even simpler if recognize set variable always as true, and unset one as false:

```go
if len(os.Getenv(name)) > 0 {
  *f = true
}
```

I prefer this way.

README.md
```diff
@@ -144,6 +144,8 @@ $ xxd -g 2 -l 64 -p /dev/random | tr -d '\n'
 * `IMGPROXY_CONCURRENCY` — the maximum number of image requests to be processed simultaneously. Default: double number of CPU cores;
 * `IMGPROXY_MAX_CLIENTS` — the maximum number of simultaneous active connections. Default: `IMGPROXY_CONCURRENCY * 5`;
 * `IMGPROXY_TTL` — duration in seconds sent in `Expires` and `Cache-Control: max-age` headers. Default: `3600` (1 hour);
+* `IMGPROXY_CACHE_FILES` — boolean value, sets whether a response sould be cached or not. If `false`, then ImgProxy will always return the status 200 and an
```
DarthSim: I don't like the name. In fact, we don't do caching at all. `IMGPROXY_USE_ETAG` would be much better.

server.go
```diff
@@ -104,11 +105,30 @@ func logResponse(status int, msg string) {
...
+			rw.WriteHeader(304)
+			logResponse(304, fmt.Sprintf("Returned cached image in %s: %s; %+v", duration, imgURL, po))
+			return
+		}
+
+	}
+
+	rw.Header().Set("Last-Modified", time.Now().Format(http.TimeFormat))
+
+	gzipped := strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") && conf.GZipCompression > 0
```
DarthSim: Please, don't move lines of code without need, even if you don't like their position.

server.go
```diff
@@ -176,8 +196,8 @@ func (h *httpHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 
 	if r.URL.Path == "/health" {
-		rw.WriteHeader(200);
-		rw.Write([]byte("imgproxy is running"));
+		rw.WriteHeader(200)
+		rw.Write([]byte("imgproxy is running"))
```
DarthSim: The same: don't make style fixes the part of feature PR.

w32blaster:
Hi, thank you for your feedback. Please check my following commit.

I decided do not generate new random value, instead, I used salt. Let me know what do you think about that. Cheers

w32blaster:
Hi, thank you for your feedback. Please check my following commit.

I decided do not generate new random value, instead I used salt. Let me know what do you think about that. Cheers

etag.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	return false
+}
+
+// function calculates the SHA checksum for the current image and current Processing Options.
+// Principal is very simple: if an original image is the same and PO are the same, then
+// the checksum must be always identical. But if PO has some different parameters, the
+// checksum must be different event if the orinal image matches
+func calculateHashSumFor(b []byte, po *processingOptions) string {
```
DarthSim: I think this way to calc ETag is not perfectly correct.

First of all, it messes image data with config, this may cause collisions.
Then it writes `po` fields one by one. Someone can add a new field to `processingOptions`, and this code will become invalid.
And it uses `conf.Salt` which keeps being constant between restarts.

What should we do:
1) Calc SHA of image data. This way image footprint will have a constant length.
2) Make all fields of `processingOptions` public. This will allow providing the whole struct to `binary.Write`.
3) Generate random cache signature at each start of imgproxy. This will reset ETag in case use cached the imgproxy config.

ETag will be calculated like this: `sha(image_footprint + po + cacheSignature)`

config.go
DarthSim: I still insist that this should be changed.

etag.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// check whether client's ETag matches current response body.
+// - if the IMGPROXY_USE_ETAG env var is unset, this function always returns false
+// - if the IMGPROXY_USE_ETAG is set, the function calculates current ETag and compare it
+//   with another ETag value provided by client
+// Note that the calculated ETag is saved to outcoming response with "ETag" header.
+func isETagMatching(b []byte, po *processingOptions, rw *http.ResponseWriter, r *http.Request) bool {
+
+	if conf.ETagEnabled {
```
DarthSim: Prefer guard clause.

```go
if !conf.ETagEnabled {
  return false
}

// calc ETag
```

And avoid redundant empty lines, please.

README.md
```diff
@@ -144,6 +144,9 @@ $ xxd -g 2 -l 64 -p /dev/random | tr -d '\n'
 * `IMGPROXY_CONCURRENCY` — the maximum number of image requests to be processed simultaneously. Default: double number of CPU cores;
 * `IMGPROXY_MAX_CLIENTS` — the maximum number of simultaneous active connections. Default: `IMGPROXY_CONCURRENCY * 5`;
 * `IMGPROXY_TTL` — duration in seconds sent in `Expires` and `Cache-Control: max-age` headers. Default: `3600` (1 hour);
+* `IMGPROXY_USE_ETAG` — boolean. Shows whether a response should be cached or not. If the variable is not presented, then ImgProxy will always return an
```
DarthSim: The description seems to be bulky and not clear. I think this would be enough:

> when set, enables using [ETag](https://en.wikipedia.org/wiki/HTTP_ETag) header for the cache control. Default: unset;

w32blaster:
Hey, thanks for your code review. Please check out two more commits with fixes. Cheers.

server.go
```diff
@@ -197,6 +199,14 @@ func (h *httpHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 
 	t.Check()
 
+	match := isETagMatching(b, &procOpt, &rw, r)
+	if match {
```
DarthSim: You don't use `match` anywhere else, you can use `isETagMatching` as condition:

```go
if isETagMatching(b, &procOpt, &rw, r) {
  ...
}
```

w32blaster:
I am sorry, have you received my questions in the code above?

etag.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"log"
+	"net/http"
+)
+
+var randomValue []byte
+
+// run on the server startup
+func init() {
+	if conf.ETagEnabled {
```
DarthSim: Are you sure this code will be executed after `conf` initialization?

DarthSim:
No, I don't see any of your questions.

etag.go
w32blaster: Well, I am afraid to say a banal phrase, but "it works on my computer" :smile:  Seriously, I run that code and I use it for my project for testing, and it works as expected, so the answer is "yes"

w32blaster: But surely, I can double check

w32blaster:
Hm... that's a shame that you didn't receive any notifications :disappointed: Ok is there anything apart of boolean argument parsing that I can do else for this PR?

DarthSim:
Three things need your attention:

1. Boolean variable parsing
2. Description of the feature in README
3. `init()` in `etag.go`

w32blaster:
Can you please answer my question under the `boolEnvConfig()` method. Thanks.

config.go
DarthSim: Well, commonly used envconfig uses `strconv.ParseBool`, so let's go this way. But it definitely should be implemented like this:

```go
if env, err := strconv.ParseBool(os.Getenv(name)); err == nil {
  *b = env
}
```

etag.go
DarthSim: Let's move it to `config.go`, and move `randomValue` to `conf`

config.go
w32blaster: Agreed :+1:

w32blaster:
Hi again,
so, I applied suggested adjustments:

* made "RandomValue" part of `conf`
* generate the random value on the configuration initialization step rather than inside `init()` function
* boolean variable parsing is implemented in the suggested way

And I think you, as the project author, can adjust Readme message about new option by yourself in any flavor you wish, it is only one sentence :+1: 

Thanks

w32blaster:
I see the build is broken in TravisCI :disappointed: with the message `go build github.com/DarthSim/imgproxy: invalid flag in #cgo LDFLAGS: -s` and here I need your help. I can't reproduce that on my computer. Can you suggest how can I fix that? Cheers!

README.md
DarthSim: `when true, enables using [ETag](https://en.wikipedia.org/wiki/HTTP_ETag) header for the cache control. Default: false;`

shutdown_old.go
```diff
@@ -1,11 +0,0 @@
...
-
-package main
-
-import (
-	"net/http"
-)
-
-func shutdownServer(_ *http.Server) {
-	// Nothing we can do here
-}
```
DarthSim: Why did you do this?

etag.go
```diff
@@ -0,0 +1,40 @@
...
+// The principal is very simple: if an original image is the same and POs are the same, then
+// the checksum must be always identical. But if PO has some different parameters, the
+// checksum must be different even if original images match
+func calculateHashSumFor(b []byte, po *processingOptions) string {
+	hash := sha1.New()
+
+	// SHA checksum consists of image, porcessing options and random value, generated on the server startup.
+	hash.Write(b)
+	binary.Write(hash, binary.LittleEndian, *po)
+	hash.Write(conf.RandomValue)
```
DarthSim: I have noticed that you ignored one of my comments:

> I think this way to calc ETag is not perfectly correct.
>
> First of all, it messes image data with config, this may cause collisions.
> ...
>
> What should we do:
>
> 1. Calc SHA of image data. This way image footprint will have a constant length.
> 2. ...
> 3. ETag will be calculated like this: sha(image_footprint + po + cacheSignature)

```go
footprint := sha1.Sum(b)

hash := sha1.New()
hash.Write(footprint[:])
binary.Write(hash, binary.LittleEndian, *po)
hash.Write(conf.RandomValue)

return fmt.Sprintf("%x", hash.Sum(nil))
```

w32blaster: why don't you change it by yourself? It is only one change of code, it is much faster and easier than writing these long comments on Github. The same question is regarding new lines and unnecessary comments deleting.

DarthSim: Well, because it's you who wants this feature, not me. I appreciate your contribution, but I don't want to merge the code that doesn't meet my requirements. Because you won't maintain the project, I will. Ok, I could fix comments and blank lines things by myself. But I definitely won't merge feature that doesn't work properly. If you corrected my comments at once, the feature would be merged already, but you prefer "why can't you do my work yourself" thing. Come on.

w32blaster: done :white_check_mark:

DarthSim:
LGTM now 👍

PR #59, opened 2018-05-26
Enable S3 protocol

crohr:
I'm currently considering replacing imgix.com with imgproxy, and was surprised to not find an option to support fetching files from S3 buckets. So I went ahead and added a way to register s3 as a transport to be able to use `s3://{BUCKET_NAME}/{OBJECT_KEY}` as a source image url.

I'm not a Go expert, so let me know if it's missing stuff!


DarthSim:
Hi Cyril,

Thanks for your contribution, and sorry for so long delay. No excuses for me here.

There are few things I'd like to fix before this PS is being merged:

1. I'm not a fan of mixing ENV variables, I like all application-related vars to have a prefix, in this case, `IMGPROXY_`. I'd like all AWS var to be named like `IMGPROXY_AWS_*`. You can use their values for creatins session like this:

  ```go
  sess, err := session.NewSession(&aws.Config{
      Region:      aws.String(config.AWSRegion),
      Credentials: credentials.NewStaticCredentials(config.AWSKey, config.AWSSecret, ""),
  })
  ```

2. AFAIK region is quite important for AWS. We defenitely should force user to specify it.

3. Since we use `IMGPROXY_AWS_*` mask for all AWS variables, we can check S3 readiness by checking their presence. There is no need in `IMGPROXY_USE_S3`.

4. I think it's better to create session inside `NewS3Transport`.

5. Don't remove `*_old.go` files. They contain code for old versions of Go.

Thanks again! Looking forward for the fixes!

steveh:
Unrelated person chiming in here, but the advantage of using `AWS_{ACCESS_KEY_ID,SECRET_ACCESS_KEY,REGION}` is that you can use the AWS SDK for authentication, which also allows reading `~/.aws/credentials`, or even not using keys at all and relying on IAM roles to authenticate the request. By using your own prefix you're forced to use access keys.

https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/#ChainProvider

crohr:
@steveh yes, exactly my thoughts when I implemented the solution.

@DarthSim sorry for the old go files removals, I will revert this.

DarthSim:
Well, this makes sense then. But we need more info in the readme about authorization methods. A user uses imgproxy, not aws-sdk-go, and they shouldn't search for info about something in third-party libraries' docs.

Another couple of comments:
1) Please format variables [here](https://github.com/DarthSim/imgproxy/pull/59/files#diff-04c6e90faac2675aa89e2176d2eec7d8R239) like i did [here](https://github.com/DarthSim/imgproxy#generating-the-url)
2) I think it would be better to put a link to `Serving files from S3` [here](https://github.com/DarthSim/imgproxy/pull/59/files#diff-04c6e90faac2675aa89e2176d2eec7d8R148) instead of describing authorization methods.

Nalum:
Any updates on this? I'm looking at getting imgproxy running and will be using it with S3 so this would be very handy for me.

If it's going nowhere I wouldn't mind taking a look at fixing any remaining issues to get it merged.

DarthSim:
This PR still waits for updates. If we won't see any updates soon, I'll release this feature in v2.0.

DarthSim:
Squashed and merged manually to the [version/2.0](https://github.com/DarthSim/imgproxy/tree/version/2.0) branch. v2.0 that will be released soon.

@crohr Thanks again for your contribution!

PR #66, opened 2018-08-07
Update PHP URL generation example

adamquaile:
String interpolation is a little easier to read and understand as there's no back-and-forth between string format and arguments. 

Matches JS version: https://github.com/DarthSim/imgproxy/blob/master/examples/signature.js#L27


DarthSim:
Thanks for the contribution!

PR #77, opened 2018-10-17
Update Elixir example

SxDx:
`Base.decode16!/2` only accepts uppercase strings as default so the elixir example does not work out of the box.


DarthSim:
Thanks for the contribution!

PR #88, opened 2018-10-30
Option to use truncated signature

printercu:
Hi, and thanks for the great tool!

I'm making some small changes to make urls more user-friendly. Here is one part to make it possible to use shorter signatures. I've not found if any vulnerabilities may be introduced with truncated signatures. It only increases chance of collision, but this is acceptable in many use cases.


crypt.go
```diff
@@ -18,14 +18,21 @@ func validatePath(token, path string) error {
...
 }
+
+func signatureFor(str string) []byte {
+	mac := hmac.New(sha256.New, conf.Key)
+	mac.Write(conf.Salt)
+	mac.Write([]byte(str))
+	expectedMAC := mac.Sum(nil)
+	if conf.SignatureSize > 0 {
+		return expectedMAC[:conf.SignatureSize]
+	} else {
```
DarthSim: We don't need `else` here since `if` block ends with `return`

crypt.go
```diff
@@ -18,14 +18,21 @@ func validatePath(token, path string) error {
...
 
 	return nil
 }
+
+func signatureFor(str string) []byte {
+	mac := hmac.New(sha256.New, conf.Key)
+	mac.Write(conf.Salt)
+	mac.Write([]byte(str))
+	expectedMAC := mac.Sum(nil)
+	if conf.SignatureSize > 0 {
```
DarthSim:
```suggestion
	if conf.SignatureSize < 256 {
```

docs/configuration.md
```diff
@@ -8,6 +8,7 @@ imgproxy allows URLs to be signed with a key and salt. This feature is disabled
 
 * `IMGPROXY_KEY`: hex-encoded key;
 * `IMGPROXY_SALT`: hex-encoded salt;
+* `IMGPROXY_SIGNATURE_SIZE`: number of bytes to use for signature (before encoding to Base64);
```
DarthSim:
```suggestion
* `IMGPROXY_SIGNATURE_SIZE`: number of bytes to use for signature before encoding to Base64. Default: 256;
```

config.go
```diff
@@ -157,6 +158,7 @@ var conf = config{
 	MaxSrcDimension:       8192,
 	MaxSrcResolution:      16800000,
 	AllowInsecure:         false,
+	SignatureSize:         0,
```
DarthSim:
```suggestion
	SignatureSize:         256,
```

printercu:
Thanks for feedback! Please advise how should validation be implemented, I've not found any validation for other params.
I've changed default SignatureSize from 0 to 32 (256 / 8) as you suggested. However I don't think it's good idea:
- it introduces magic constant (or requires global constant),
- it requires more validation,
- it's harder to extend in future. Ex., introduce option to select signature algorithms.

DarthSim:
> Please advise how should validation be implemented, I've not found any validation for other params.

Like this: https://github.com/DarthSim/imgproxy/blob/master/config.go#L263-L265

> I've changed default SignatureSize from 0 to 32 (256 / 8) as you suggested. However I don't think it's good idea:
> 
> * it introduces magic constant (or requires global constant),

32 bytes (256 bits) is not a "magic constant", it is length of SHA256 hashsum.

> * it requires more validation,

It doesn't. You just need to check if config value is between 1 and 32. 

> * it's harder to extend in future. Ex., introduce option to select signature algorithms.

There won't be that kind of option.

printercu:
Added validation.

crypt_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestSignatureFor(t *testing.T) {
+	base64Signature := func(x string) string { return base64.RawURLEncoding.EncodeToString(signatureFor(x)) }
+	conf.Key = []byte("test-key")
+	conf.Salt = []byte("test-salt")
+	assert.Equal(t, "dtLwhdnPPiu_epMl1LrzheLpvHas-4mwvY6L3Z8WwlY", base64Signature("asd"))
+	assert.Equal(t, "8x1xvzxVqZ3Uz3kEC8gVvBfU0dfU1vKv0Gho8m3Ysgw", base64Signature("qwe"))
+	conf.SignatureSize = 8
+	assert.Equal(t, "dtLwhdnPPis", base64Signature("asd"))
+	assert.Equal(t, "8x1xvzxVqZ0", base64Signature("qwe"))
+	conf.SignatureSize = 0
```
DarthSim:
```suggestion
	conf.SignatureSize = 32
```

printercu:
Done

DarthSim:
Thanks for the contribution! Merged to v2.1 which will be released soon.

PR #94, opened 2018-11-01
adds support for client hints width/viewport-width.

selul:


DarthSim:
LGTM except for a couple of things:

* Please, use `version/2.1` branch as a base. I switched back from fasthttp to the built-in net/http, and your code became incompatible.
* It's better to disable client hints by default since it's a dangerous thing as I said before. Let's add a config value `IMGPROXY_ENABLE_CLIENT_HINTS`.

selul:
Thanks for the feedback, I missed the fasthttp switch, I've opened a new PR from `version/2.1` branch with the env variable too. 

https://github.com/DarthSim/imgproxy/pull/98

PR #95, opened 2018-11-01
Option to use plain source urls

printercu:
Hi! This is second pr to make urls more user-friendly.

It introduces `PlainSourceURL ` option which disables base 64 decoding of source url. To avoid double extensions in urls (`.jpg.jpg` or `.png.jpg`) I've decided to move source extension in front of urls.

Here are examples of urls (with truncated signature from #88):
```
with new option disabled: 
/QsqUhBDJtdU/fit/64/64/ce/1/L2Fzc2V0cy9pbWFnZXMvMS8yLzMucG5n.jpg 

enabled:
/EPjBZdh5zf4/fit/64/64/ce/1/png/assets/images/1/2/3.jpg 
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -143,6 +143,8 @@ type config struct {
 
 	BaseURL string
 
+	PlainSourceURL bool
+
 	Presets presets
 }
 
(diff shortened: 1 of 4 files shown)


DarthSim:
Hi,

Sorry, but I'm going to close this PR. First of all, I'm working on the same feature but in a different way. Also, I don't really like the way you choose:

1. Extensions thing really puzzles me. And URL generation becomes way more complex.
2. It doesn't work with a query string.
3. It breaks advanced URL format when you try to use full URL.

PR #97, opened 2018-11-02
Validate url format after adding BASE_URL

printercu:
This makes it possible to use `BASE_URL=http://host/path_part_` and image url `_second_part/file.ext`.

### DIFF
--- a/download.go
+++ b/download.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


DarthSim:
Oh, thanks for catching this. We don't actually need to check url on the download stage since it will be checked by `downloadClient.Get`. Please, remove the redundant check, and I'll merge this.

DarthSim:
Nevermind, I'll do it myself. Thanks for your contribution!

DarthSim:
Released in `darthsim/imgproxy:v2.0.3`

Issue #92, opened 2018-11-01
support for client hints ? 


selul:
Hi, 

Do you think will be useful to integrate Client hints https://developers.google.com/web/updates/2015/09/automating-resource-selection-with-client-hints into imgproxy ? 

Something like when the width&height is missing from the url, processing options will consider the `Width` header, or if this is not present, `Viewport-Width` header? 

Related to this it might be useful to have a DPR processing option ? 

I've been able to implement those into optimole.com, which is using a slightly modified version of imgproxy, i would be happy to contribute if you need any help. 

Thanks a lot for your hard work !

DarthSim:
Hi,

I think about headers as a dangerous territory since they can't be signed. An attacker will be able to spam your imgproxy by changing `Viewport-Width` header. They also can eat all your memory by sending a huge number in `Viewport-Width`. How do you deal with it in OptiMole?

Regarding DPR option, how is it different from just requesting a bigger size?

selul:
Thanks for the feedback, @DarthSim.

Indeed is on the gray area but when the URL is unsigned, wont make any difference i think, as I was using those headers as defaults in the processing option func and the same normalizations will apply during the process as the width used as an argument in the URL.

Regarding DPR transformation I find it handy when you want to multiply the resulting image size by a certain factor. 

Moreover, when you don't know what size the resulting image will have, i.e width/height is not passed as arguments, this is very useful to format images for HiDPI devices.

selul:
I've just added a PR https://github.com/DarthSim/imgproxy/pull/94, I would love to hear your thoughts.

DarthSim:
Done with #98

PR #104, opened 2018-11-13
adds dpr transformation 

selul:
Following the conversation from #92 I've tried to implement the DPR transformation as well as support this in client hints. 

The DPR factor can be a float number between 1 and 8 and will multiply the resulting image sizes by this factor. 
 
Looking forward to your feedback, @DarthSim


process.go
```diff
@@ -245,7 +245,19 @@ func processImage(ctx context.Context) ([]byte, error) {
 	checkTimeout(ctx)
 
 	imgWidth, imgHeight, angle, flip := extractMeta(img)
+	if po.Dpr > 1 {
+		if po.Width != 0 {
+			po.Width = int(float32(po.Width) * po.Dpr)
+		} else {
+			po.Width = int(float32(imgWidth) * po.Dpr)
```
DarthSim: This breaks the behavior. Zero width/height should stay zero.

processing_options.go
```diff
@@ -470,6 +472,19 @@ func applyBlurOption(po *processingOptions, args []string) error {
 	return nil
 }
 
+func applyDprOption(po *processingOptions, args []string) error {
+	if len(args) > 1 {
+		return fmt.Errorf("Invalid dpr arguments: %v", args)
+	}
+
+	if d, err := strconv.ParseFloat(args[0], 32); err == nil || (d >= 1 || d <= 8) {
```
DarthSim: I don't think we need the upper limit here. If a user wants to generate an image with 100500 DPR, well, as we use to say in Russia, he is his own evil Pinocchio. I think this limit is reasonable only for DPR header.

selul: hehe, makes sense.

process.go
selul: Thanks for the feedback, but what about when the users do not request any width/height in specific, and he just wants to get an image with dpr=2 for instance. where is the right location to apply this?

selul:
Hey, 

I was thinking about the DPR transformation something that should apply to the resulting image, indifferent of what kind of transformation was done before.  
Furthermore, if the image is cropped to 200x100 after the cropping is done and the resulting image is ready, we should rescale the image according to the DPR factor. Looking at some cloud-based image services, it seems they follow the same logic: 
https://docs.imgix.com/apis/url/dpr 

Thinking more about this, I might not have added the dpr logic in the proper place in `process.go` ? What do you think?

DarthSim:
Yeah, I think we need more complex behavior for DPI. Let's keep it as is in this PR (with fixes), and I'll think how to make it properly.

processing_options.go
DarthSim: Also, i think it's a good idea to support LowDPI, so let's allow dpi < 1 and > 0.

DarthSim:
Thanks for the contribution again!

PR #119, opened 2018-12-20
Vary: Accept response header

leopangan:
For the Accept request header to be considered by Google Cloud CDN, the response must have Vary: Accept header. See https://cloud.google.com/cdn/docs/caching#vary_headers


DarthSim:
Hi!

Thanks for the contribution! Could you also add client hints headers to `Vary` when ones are enabled?

leopangan:
Sure @DarthSim! I guess we should also add Accept-Encoding when GZipCompression is enabled?

DarthSim:
Yeah, looks like we should, thanks

leopangan:
@DarthSim done

DarthSim:
Looks great, thanks!

PR #128, opened 2019-01-11
Add Sentry error reporting

### LINKED ISSUE #125: Support for logging errors to Sentry
As an alternative to New Relic it would be nice to also support Sentry. 

I could take a stab at implementing this, but I'm not that experienced with Go, and I currently probably don't have time for it..

Documentation: https://docs.sentry.io/clients/go/

koenpunt:
This PR adds support for logging error to Sentry.

~The Sentry client already supports setup using environment variables (when the library is included a default client is created using them), but I'm not sure if we want to rely on that from within imgproxy, so I created a dedicated client instead.~

~Alternative would be to use the default client, but configure it using the imgproxy config.~

Nevermind, I'm using the default client now.

Fixes #125


DarthSim:
Looks great to me, thanks!

Please, fix the link to the error reporting section. It became broken when you changed the header

koenpunt:
Done!

DarthSim:
Merged. Thanks for your contribution again!

PR #130, opened 2019-01-15
move healthcheck before the secret checking

koenpunt:
Since health checks on AWS target groups do not support sending headers, it's impossible to use the health check when a secret is configured.

I think the health won't cost more resources than the panic, if not less, so I think it's safe to move this before the secret check.


DarthSim:
Makes sense. Thanks!

Issue #115, opened 2018-12-15
Add examples for truncated urls


brandonkal:
I noticed truncated signatures are now supported. I've attempted to to create a working hash but imgproxy is returning 403 invalid signature.

Here is what I have now:
```
/**
 * Hash image url for imgproxy use
 */
function eml_make_image_url( $img_path, $w = 0, $h = 0, $type = 'fill', $png = false ) {
	$key_bin = pack( 'H*', IMGPROXY_KEY );
	$salt_bin = pack( 'H*', IMGPROXY_SALT );
	if ( empty( $key_bin ) || empty( $salt_bin ) ) {
		die( 'Key/Salt expected to be hex-encoded string' );
	}
	// if image is a custom-sized png or logo, allow png extension
	$ext = ( $png && substr( strtolower( $img_path ), -4 ) === '.png' ) ? '/f:png' : '';

	$encoded_url = rtrim( strtr( base64_encode( 'local:///' . $img_path ), '+/', '-_' ), '=' );
	$path = "/rs:{$type}:{$w}:{$h}{$ext}/{$encoded_url}";
	$hash = hash_hmac( 'sha256', $salt_bin . $path, $key_bin, true );
	// Truncate the hash to 8 bytes before encode
	$hash = mb_strcut( $hash, 0, 8, 'UTF-8' );
	$signature = rtrim( strtr( base64_encode( $hash ), '+/', '-_' ), '=' );
	return BASE_URL . '/img/' . $signature . $path;
}
```

I also have `IMGPROXY_SIGNATURE_SIZE: 8`.

Hoping you can shed some light on this and possible update the examples. Thanks!

https://github.com/DarthSim/imgproxy/blob/280c64a55b170afcf297c56d8e17e78d632f507c/examples/signature.php#L28

selul:
@brandonkal I've recently done this, and I've used `pack` to truncate the signature, here is a working example: 
```
		$binary         = hash_hmac( 'sha256', $salt.$path , $key, true );
		$binary         = pack("A8", $binary); //replace 8 with the signature size 
		$$signature= rtrim( strtr( base64_encode( $binary ), '+/', '-_' ), '=' );
```

DarthSim:
Great! Thanks, @selul!

If you guys think this should be included in the example, please, make a PR.

PR #131, opened 2019-01-17
adds truncated signature example for PHP

selul:
following #115 i've added an example on how to use signature size for php


DarthSim:
I think the truncated signature example should be located in the separate file to not puzzle new users.

selul:
@DarthSim makes sense. I've added a separate file.

DarthSim:
Great, thanks!

PR #134, opened 2019-02-04
Update installation.md

charlie-wasp:
`vips` was migrated to `homebrew/core`, so it's not necessary to use `science` tap anymore


DarthSim:
👍

PR #156, opened 2019-04-30
Adds an option to allows only presets in the URL.

carlosbrando:
If you set `IMGPROXY_ONLY_PRESETS` as `true`, a preset is obligatory, and all other URL formats are disabled.

In this case, you always need to inform a preset in your URLs without the `preset` or `pr` statement. Example: `http://imgproxy.example.com/AfrOrF3gWeDA6VOlDG4TzxMv39O7MXnF4CXpKUwGqRM/thumbnail/aHR0cDovL2V4YW1w/bGUuY29tL2ltYWdl/cy9jdXJpb3NpdHku/anBn.png`

It's possible to use more than one preset separing them with `:` like `thumbnail:gray`.

### DIFF
--- a/config.go
+++ b/config.go
@@ -180,7 +180,8 @@ type config struct {
 
 	BaseURL string
 
-	Presets presets
+	Presets     presets
+	OnlyPresets bool
 
 	WatermarkData    string
 	WatermarkPath    string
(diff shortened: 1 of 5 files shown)


carlosbrando:
I want to give some context to this pull request: Here we didn't want to allow our users to define their image formats. So we are always using presets for that, and we want to guarantee that only presets are accepted.

DarthSim:
Hi,

Patching `parseURLOptions` breaks parsing of presets as I see. It's better to add function `parsePathPresets` which will act like `parsePathAdvanced` and `parsePathBasic`.

carlosbrando:
@DarthSim can you please check it again?

DarthSim:
Looks perfect. Will release it soon. Thanks for your contribution!

PR #158, opened 2019-05-01
Fixes dockerfile to contain fonts

### LINKED ISSUE #157: SVG and supported fonts
I have been trying to use the SVG -> PNG conversion. It seems like something is not installed correctly in the docker image for the fonts to work. 

Here is my original SVG:

Screen Shot 2019-05-01 at 7 43 55 AM

Here is that same SVG in JPEG format:

Screen Shot 2019-05-01 at 7 44 15 AM

Fonts are all broken. I have played around with CairoSVG and I remember true fonts had to be installed. Has this been an issue for anyone else? 

If you point me to the right direction, I can provide a PR. 

PS. This is an amazing project. I have been looking for something like this for a long time! The webp auto detection is the best.

amir20:
This fixes #157 by adding fontconfig to Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,6 +77,45 @@ RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/reposit
   && apk --no-cache upgrade \
   && apk add --no-cache bash ca-certificates fftw glib expat libjpeg-turbo libpng \
     libwebp giflib librsvg libgsf libexif lcms2 libimagequant\
+  && apk add --no-cache \
+      fontconfig \
+      font-bh-100dpi \
+      font-sun-misc \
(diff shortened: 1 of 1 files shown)


amir20:
See example below: 

Screen Shot 2019-05-01 at 1 04 24 PM

amir20:
The only downside is that it adds about 40MB to the docker image.

amir20:
The fonts are very jagged. I haven't been able to fix it. I think since it adds 40MB and it's not perfect, people should extend the dockerimage to do their own thing. That's just my $.02

DarthSim:
Yeah, I agree. 40 MB is too much for such an uncommon feature. Since it's possible to extend the base image with additional packages, I don't think we should add fonts to the base image.

PR #159, opened 2019-05-02
"Invalid source image" hides important, readily available, relevant diagnostic information

rdm:
When imgproxy rejects a valid jpeg source image, it's good to know something about its reason for emitting that error.

### DIFF
--- a/download.go
+++ b/download.go
@@ -24,10 +24,10 @@ var (
 	imageTypeCtxKey = ctxKey("imageType")
 	imageDataCtxKey = ctxKey("imageData")
 
-	errSourceDimensionsTooBig      = newError(422, "Source image dimensions are too big", "Invalid source image")
-	errSourceResolutionTooBig      = newError(422, "Source image resolution is too big", "Invalid source image")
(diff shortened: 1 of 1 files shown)


rdm:
I had two issues setting up an imgproxy instance:

(1) The imgproxy default resolution was too low to handle the images I was working with. The error message seemed to imply that the image itself was invalid. This pull request addresses that issue.

(2) I was working with a code base which only handled the basic url format and was trying to do a resize operation. The error message I got did not hint at this issue. Making the error message for this case informative seems to indicate refactoring processing_options.go. I am still thinking about what would be best for this issue.

DarthSim:
Hi,

This was done intentionally.  I'm a bit of paranoid in cases of sharing internal info with the world, that's why imgproxy doesn't show full error info in the response. On the other hand, logs contain all the info you want. v2.2.13+ sends `X-Request-ID`, so you can easily grep your logs for the specific request. v2.3 includes "development errors mode" which can be enabled with `IMGPROXY_DEVELOPMENT_ERRORS_MODE`. In this mode, imgproxy responds with the same error message as it writes to log.

PR #163, opened 2019-05-25
Add javascript example for truncated signature

koenpunt:
Since some of the other examples also demonstrate the use truncated signatures, I've added it for the javascript example.

### DIFF
--- a/examples/signature-truncated.mjs
+++ b/examples/signature-truncated.mjs
@@ -1,0 +1,21 @@
+import { createHmac } from 'node:crypto';
+
+const KEY = '943b421c9eb07c830af81030552c86009268de4e532ba2ee2eab8247c6da0881'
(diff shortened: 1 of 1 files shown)


DarthSim:
Hi,

Thanks for your contribution! Could you put the truncated signature example in a separate file so it won't puzzle ones who don't need truncation?

koenpunt:
> Could you put the truncated signature example in a separate file

Ha, I saw a similar comment on the PHP version after I submitted this PR, so I was already expecting this feedback. Although I think in this case it's a bit different, since it's just a simple slice on the digest value, whereas with the PHP version there's a new method call introduced. 

But if you really want to me to split it I can do that as well.

DarthSim:
I like examples to be copy-paste ready with as less redundant code as possible. So let's split it :)

koenpunt:
Took me some time, but I updated this PR to have the example in a separate file.

DarthSim:
Thanks for your contribution!

Issue #167, opened 2019-06-15
Disable non-local sources


rhyst:
Is it possible to disable non-local file sources? I.e I only want imgproxy to serve images from my server, and no where else.

Thanks.

DarthSim:
Hi,

The easiest way is to set `IMGPROXY_BASE_URL=local:///` and provide source URLs without schema.

DarthSim:
Closed with #169

PR #169, opened 2019-06-16
Add config option to exclusively allow local file system images

rhyst:
From my issue: #167 

Allows you to restrict sources to local file system only.

I've not written go before so I apologise if this is awful.


DarthSim:
Hi,

Thanks for your contribution!

Let's change a couple of things:

1. Let's move schema checking to the processing options parsing stage.
2. I don't think it's a good idea to add a restriction specifically for a `local` scheme. Let's go with `IMGPROXY_ALLOWED_SOURCES` option. When it's empty, all schemas are allowed. When it's set to something like `https,local,s3`, only the listed schemas are allowed.

rhyst:
Hey. I had a go at the changes you suggested. If it looks good I assume I should add this to the documentation. What do you think is the most appropriate page?

processing_options.go
```diff
@@ -763,6 +764,19 @@ func applyProcessingOptions(po *processingOptions, options urlOptions) error {
...
+	if conf.AllowedSources == "" {
+		return true
+	}
+	for _, val := range strings.Split(conf.AllowedSources, ",") {
+		if string(val) == source {
+			return true
+		}
+	}
+	return false
+}
```
DarthSim: There are a few problems with this code:

1) `source := strings.Split(imageURL, ":")[0]` creates a new slice of strings on every request, which leads to redundant memory allocations. More of all, it will cause panic if `imageURL` doesn't include `:`. Using `strings.HasPrefix` is much more efficient way.
2) `strings.Split(conf.AllowedSources, ",")` creates a new slice of strings on every request as well. It's better to implement a custom config parser that will convert `IMGPROXY_ALLOWED_SOURCES` value from something like `local,s3,gs` to something like `[]string{"local://", "s3://", "gs://"}`, and use this parser on the configuration stage.

download.go
```diff
@@ -173,6 +174,11 @@ func requestImage(imageURL string) (*http.Response, error) {
 func downloadImage(ctx context.Context) (context.Context, context.CancelFunc, error) {
 	imageURL := getImageURL(ctx)
 
+	_, err := neturl.Parse(imageURL)
+	if err != nil {
+		return ctx, func() {}, newError(404, err.Error(), msgSourceImageIsUnreachable)
+	}
+
```
DarthSim: This is a redundant code

processing_options.go
```diff
@@ -952,6 +966,10 @@ func parsePath(ctx context.Context, r *http.Request) (context.Context, error) {
 		imageURL, po, err = parsePathAdvanced(parts[1:], headers)
 	}
 
+	if !isAllowedSource(imageURL) {
+		return ctx, newError(404, fmt.Sprintf("Invalid source"), msgInvalidSource)
+	}
+
```
DarthSim: This should be placed after the following error check

rhyst: Removed

processing_options.go
rhyst: Implemented custom parser and switched to using HasPrefix

download.go
rhyst: Removed :)

rhyst:
Another go :grin:

DarthSim:
LGTM now! Thanks for your contribution. Don't forget to remove debug warnings in the future, please.

DarthSim:
I changed `IMGPROXY_ALLOWED_SOURCES` behavior a bit so that it accepts URL prefixes instead of protocols. This allows setting more complex limitations like `local://,https://example.com,https://images.com/photos`

PR #172, opened 2019-06-19
Add link to imgproxy package in AUR for Arch Linux users

Envek:
Fun fact: actually AUR doesn't contain packages and software is being build every time from the sources on end user's machine. So it is long and loud process. But it is as simple as package manager – you just check imgproxy and it will be installed for you along with libvips.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -1,6 +1,6 @@
 # Installation
 
-There are three ways you can install imgproxy:
+There are four ways you can install imgproxy:
 
 ### Docker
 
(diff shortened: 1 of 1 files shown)


DarthSim:
Thanks! Looks like it's time to upgrade the package to v2.3 ;)

Envek:
> Looks like it's time to upgrade the package to v2.3 ;)

Oops, I missed the release notification from GitHub. [Done!](https://aur.archlinux.org/cgit/aur.git/commit/?h=imgproxy&id=5286fbbe63bf8d281235017b7ef90f131582d80e)

PR #176, opened 2019-07-11
Add BMP support

dmitryzuev:


DarthSim:
Hi,

Thanks for your contribution!

I try to avoid `magickload` since ImageMagick is dark and full of terrors. Also, I'm pretty sure BMP is dead, so I don't think imgproxy needs its support.

I'll keep this PR open to see if there will be upvotes.

DarthSim:
I finally decided to use `magickload` for certain formats.

Thanks for your contribution!

PR #177, opened 2019-07-12
Add TIFF support

dmitryzuev:
I've tried to use `golang.org/x/image/tiff` lib, but it was unable to decode CMYK tiff files.
And I decided to copy `DecodeConfig`, but use it only to detect width and height of remote file.

Probably `tiff.go` need a little bit of refactoring


DarthSim:
Hi,

Thanks for your contribution!

`tiff.go` contains a lot of ambiguous names like `decode`, `length`, `buffer`, `Decode` and so on. I think it's better to move it to a subpackage.

DarthSim:
Also, I would like to get rid of the `buffer` type. The only line that stops us from doing that is https://github.com/imgproxy/imgproxy/pull/177/files#diff-f589ead8c094014ca8c8d7e6df51889eR122, but we could drop support of pointers since width/height most likely will never be encoded with something bigger than 4 bytes.

If we remove `buffer`, we can just rename constants, types, and functions without moving them to a subpackage.

DarthSim:
I implemented TIFF support properly in https://github.com/imgproxy/imgproxy/commit/5e18d59f809799388c9f745475f8a215f23ff582

Thanks for your contribution anyway!

dmitryzuev:
Thank you!

PR #179, opened 2019-07-17
Docker support for HEIC

mcfedr:
Currently support is broken because libheif is not installed.

I've also moved the `ADD` command to after the deps build so they don't have to be rebuilt for every change to go files.


DarthSim:
Hi,

Thanks for your contribution! I was somehow sure that I've added libheif to the image.

> I've also moved the ADD command to after the deps build so they don't have to be rebuilt for every change to go files.

This was done intentionally to build new versions with the latest packages versions. But `--no-cache`  is a better solution for this, for sure.

PR #181, opened 2019-07-23
Use VIPS_META_ORIENTATION instead of EXIF orientation

mcfedr:
This is set for all image formats, including hief which
dont use EXIF but can be rotated

### DIFF
--- a/vips.c
+++ b/vips.c
@@ -163,12 +163,11 @@ vips_heifload_go(void *buf, size_t len, VipsImage **out) {
 
 int
 vips_get_exif_orientation(VipsImage *image) {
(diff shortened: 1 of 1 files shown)


DarthSim:
Hi,

`VIPS_META_ORIENTATION` is supported only since libvips 8.8, so we need a couple of `ifdef`-ed parts of code for old and new versions. I decided not to do this at some point to not complex the code, but it looks like we have to do it.

DarthSim:
Done properly with https://github.com/imgproxy/imgproxy/commit/447e53fc4ea40a4346f7429827564b6863c40978

PR #188, opened 2019-08-09
Fix config.go

a1ien:
Without this i got
./config.go:461:41: constant 4294967295 overflows int
./config.go:467:37: constant 4294967295 overflows int


DarthSim:
This commit should fix it https://github.com/imgproxy/imgproxy/commit/bf64a4878330d533d1b944066a75824edd1b4e0d

PR #189, opened 2019-08-09
Fix build for 32bit platform

a1ien:
Better we need use for 64 bit 6GB but for 32bit 2GB.
But I not know go


DarthSim:
This commit should fix it https://github.com/imgproxy/imgproxy/commit/bf64a4878330d533d1b944066a75824edd1b4e0d

PR #200, opened 2019-09-04
Fixed browser usage

micheleriva:
Using `const createHmac = require('create-hmac')` will ensure that this script will work both on Node.js and Browser (otherwise it will crash on browser).
I've also updated `new Buffer` on line 7, which is deprecated in flavor of `Buffer.from`


DarthSim:
Hi,

Signing URL in a browser isn't secure since you're leaking your keys. But why not.

Thanks for your contribution!

PR #275, opened 2019-11-27
Implement Max Bytes Filter

dmitryzuev:
In some cases I needed to get image with size less than specified (e.g to pass some custom RSS feeds specifications).

#### Max Bytes

```
max_bytes:%max_bytes
mb:%max_bytes
```

This filter automatically degrades the quality of the image until the image is under the specified amount of bytes.

*Warning: this filter processes image multiple times to achieve specified image size*

Default: 0

### DIFF
--- a/process.go
+++ b/process.go
@@ -133,6 +133,15 @@ func canScaleOnLoad(imgtype imageType, scale float64) bool {
 	return imgtype == imageTypeJPEG || imgtype == imageTypeWEBP
 }
 
+func canFitToBytes(imgtype imageType) bool {
+	switch imgtype {
+	case imageTypeJPEG, imageTypeWEBP, imageTypeHEIC, imageTypeTIFF:
+		return true
+	default:
+		return false
+	}
+}
+
 func calcJpegShink(scale float64, imgtype imageType) int {
 	shrink := int(1.0 / scale)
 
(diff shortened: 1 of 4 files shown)


DarthSim:
Hi,

Thanks for your contribution!

Unfortunately, your solution is not optimal and needs a rework:

1. You don't need to rerun the whole processing flow since quality affects only image saving. All you should do is to call your quality-trimming wrapper instead of `img.Save(...)` at the end of `processImage` if `po.MaxBytes` isn't zero.
2. If you aren't satisfied with the result, you should immediately call cancel function returned by `img.Save()`. The result buffer isn't garbage collected because it's allocated by C. `cancel` frees this buffer, and if you delay its call, you can run into OOM easily.
3. It would be good to call `img.CopyMemory()` at the beginning of your quality-trimming wrapper. This function will prepare the image pixels, so `img.Save()` won't need to do in at each try.

dmitryzuev:
@DarthSim thanks for comment!

I'll try to rewrite it

process.go
```diff
@@ -682,5 +691,52 @@ func processImage(ctx context.Context) ([]byte, context.CancelFunc, error) {
 		checkTimeout(ctx)
 	}
 
+	if po.MaxBytes > 0 && canFitToBytes(po.Format) {
+		return processToFitBytes(ctx, img)
```
DarthSim: You can use `po` instead of `ctx` here

process.go
```diff
@@ -682,5 +691,52 @@ func processImage(ctx context.Context) ([]byte, context.CancelFunc, error) {
...
+			diff = 0.75
+		}
+		quality = int(float64(quality) * diff)
+		if quality < 10 {
+			return initialResult, func() {}, nil
+		}
+
+		checkTimeout(ctx)
+
+		img.CopyMemory()
```
DarthSim: Don't need to copy memory before each save. It should be done once at the beginning of this func.

process.go
```diff
@@ -682,5 +691,52 @@ func processImage(ctx context.Context) ([]byte, context.CancelFunc, error) {
...
+		case delta > 3:
+			diff = 0.25
+		case delta > 1.5:
+			diff = 0.5
+		default:
+			diff = 0.75
+		}
+		quality = int(float64(quality) * diff)
+		if quality < 10 {
+			return initialResult, func() {}, nil
```
DarthSim: I think it would be better to return the last result

DarthSim: I see the whole function like this:

```go
img.CopyMemory()

quality := po.Quality

for {
	result, cancel, err := img.Save(po.Format, quality)
	if len(result) <= po.MaxBytes || quality <= 10 || err != nil {
			return result, cancel, err
	}

	cancel()

	// trim quality
}
```

DarthSim:
LGTM now. Thanks again for your contribution!

PR #284, opened 2019-12-05
Add Homebrew package to installation methods

renchap:
`imgproxy` formula has been merged into Homebrew: https://github.com/Homebrew/homebrew-core/pull/47489

This will make this easier for development environments on macOS when not using docker, as you no longer need to have `go` installed, and upgrades will be managed by `brew`.


DarthSim:
Hi!
Thanks a lot for the formula!

PR #300, opened 2019-12-17
fix(docs): low fps on scroll on safari

euaaaio:


DarthSim:
Thanks for your contribution!

PR #313, opened 2019-12-26
Use image height to calculate watermark top offset

realsim:
Shouldn't we use image height to properly calculate watermark top offset?


DarthSim:
Oops! Thanks for the fix!

PR #328, opened 2020-01-21
Don't scale the crop gravity's X and Y

### LINKED ISSUE #327: Crop gravity should not be scaled when resizing
I get wrong results when using requests like this one

    /unsigned/crop:800:600/gravity:fp:1:1/resize:fill:400:300/plain/https://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_0.jpg

where [the image](https://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_0.jpg) is 1200x1800px. I would expect imgproxy to first crop the lower right 800x600px, then resize them to 400x300px (again, if anything, gravitating towards the lower right). Instead, it returns a cropped 400x300px area from the center of the image.

I traced this back to https://github.com/imgproxy/imgproxy/blob/ee450bec469d31f51f04b0929a998bf8dcfbe292/process.go#L327 and the line after that, where the crop gravity's `X` and `Y` are scaled according to the resize scaling factor (even though they are relative numbers). Removing those lines returned the expected result.

I'm not entirely sure why my example returns a center crop, to be honest, especially when comparing it to

    /unsigned/gravity:fp:1:1/resize:fill:400:300/plain/https://raw.githubusercontent.com/recurser/exif-orientation-examples/master/Portrait_0.jpg

which returns exactly the bottom half (correctly, by the way).

Even though the docs say cropping is done before resizing the image, the code smartly works around that, I assume in order to make use of scaled loading. So in reality, images are resized first and cropped afterwards. But then why does the query with cropping return pixels that are missing from the result I get when I leave out the cropping?

I'm not entirely sure whether my fix breaks something else, but at first thought, it seems wrong to scale relative gravity offsets.

sauerbraten:
when both cropping and resizing is requested

fixes #327

### DIFF
--- a/process.go
+++ b/process.go
@@ -324,8 +324,6 @@ func transformImage(ctx context.Context, img *vipsImage, data []byte, po *proces
 
 	cropWidth = scaleInt(cropWidth, scale)
 	cropHeight = scaleInt(cropHeight, scale)
-	cropGravity.X *= scale
-	cropGravity.Y *= scale
 
 	if !trimmed && scale != 1 && data != nil && canScaleOnLoad(imgtype, scale) {
 		if imgtype == imageTypeWEBP || imgtype == imageTypeSVG {


DarthSim:
I fixed this with https://github.com/imgproxy/imgproxy/commit/03bc5c820ccf1286c1604b5e8a4ff27fa1373b4a. X/Y should be scaled unless gravity type is not `fp`.
Thanks for your contribution anyway!

Issue #325, opened 2020-01-20
Offer to keep image metadata


sauerbraten:
Right now, it seems imgproxy strips all metadata when saving JPEG images at the end of the pipeline: https://github.com/imgproxy/imgproxy/blob/ee450bec469d31f51f04b0929a998bf8dcfbe292/vips.c#L516

Is there an option to keep tags? If not, should I send a PR that adds this as an env var?

sauerbraten:
Implemented in #329.

PR #329, opened 2020-01-21
Add option to keep metadata/tags in output image

sauerbraten:
When `IMGPROXY_STRIP_METADATA=false`, the vips save functions for JPEG and WebP are called with `(..., "strip", FALSE, ...)`. Image rotation metadata is removed when calling vipsImage's Rotate method.

Implements #325.

### DIFF
--- a/vips.go
+++ b/vips.go
@@ -130,6 +130,13 @@ func vipsLoadWatermark() (err error) {
 	return
 }
 
+func gbool(b bool) C.gboolean {
+	if b {
+		return C.gboolean(1)
+	}
+	return C.gboolean(0)
+}
+
 func (img *vipsImage) Width() int {
 	return int(img.VipsImage.Xsize)
 }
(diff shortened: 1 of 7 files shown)


vips.go
```diff
@@ -312,6 +319,8 @@ func (img *vipsImage) Rotate(angle int) error {
 		return vipsError()
 	}
 
+	C.vips_autorot_remove_angle(tmp)
```
sauerbraten: Is this the right place to strip orientation tags? I also thought about doing it when fetching the image orientation during extractMeta(), but then a second call to extractMeta() would return different values.

Another place where I think this would make sense is vipsImage.Save().

rufoa:
Would be great if this could be merged!

DarthSim:
Hi,

Looks good overall, but we need to remove an ICC profile from the image since we may not remove it. You can do this by modifying `vips_icc_import_go`:

```c
int
vips_icc_import_go(VipsImage *in, VipsImage **out, char *profile) {
  if (vips_icc_import(in, out, "input_profile", profile, "embedded", TRUE, "pcs", VIPS_PCS_XYZ, NULL))
    return 1;

  vips_image_remove(*out, VIPS_META_ICC_NAME);

  return 0;
}
```

DarthSim:
Thanks for your contribution!

PR #357, opened 2020-03-06
Improve trim method to accept color to be cut off and hor/ver alignment options

a1tus:


DarthSim:
Hi,

The PR looks very good. Please make a few things, and I'll merge it:
- Make all arguments after `threshold` optional. Just check that they're not empty before parsing. You can look at [applyWatermarkOption](https://github.com/imgproxy/imgproxy/blob/master/processing_options.go#L687) as an example.
- Make `Smart` true by default. Make it false if a user sets a custom color.
- Add documentation for the new arguments.

a1tus:
> Make all arguments after threshold optional. Just check that they're not empty before parsing. You can look at applyWatermarkOption as an example.
It seems to me that they are optional now: every check in `applyTrimOption` has a check `if nArgs > N`.

> Make Smart true by default. Make it false if a user sets a custom color.
My bad, it's a bug after final refactor :(

> Add documentation for the new arguments.
Will do.

Also, I hope you've checked C code thoroughly (although it's quit simple).

DarthSim:
> It seems to me that they are optional now: every check in applyTrimOption has a check if nArgs > N.

This only checks that argument doesn't exist but doesn't check if it's empty. For example, it's currently not possible to do something like this `trim:::1` (skipping the first two args).

> Also, I hope you've checked C code thoroughly (although it's quite simple).

Yep, LGTM

a1tus:
All done.

DarthSim:
Thanks for your contribution again!

PR #358, opened 2020-03-06
Implement padding option

a1tus:


DarthSim:
Hi,

PR looks good overall. Please make a few things, and I'll merge it:

- Move the padding code right before the extending code.
- Remove the color argument. I think it's better to use background color as extending does.
- Make the other arguments more CSS-like. The first one sets all the dimensions, the second one sets left and right, the third one sets bottom, the 4th one sets left.
- Add documentation for the new option.

a1tus:
> Move the padding code right before the extending code.

Not quite understand. You mean code in `newProcessingOptions` and `applyPaddingOption` or actual `embed` func call?

> Remove the color argument. I think it's better to use background color as extending does.

Was thinking about it while writing code but decided not to clash them at first iteration.

> Make the other arguments more CSS-like. The first one sets all the dimensions, the second one sets left and right, the third one sets bottom, the 4th one sets left.

Was thinking about it too :)

> Add documentation for the new option.

Will do

DarthSim:
> Not quite understand. You mean code in newProcessingOptions and applyPaddingOption or actual embed func call?

Sorry for puzzling. I mean put [this](https://github.com/imgproxy/imgproxy/blob/2971f533dc50ce6fcc2d202025c8a6416db477f9/process.go#L393-L401) => [here](https://github.com/imgproxy/imgproxy/blob/2971f533dc50ce6fcc2d202025c8a6416db477f9/process.go#L464).

> Was thinking about it while writing code but decided not to clash them at first iteration.

Let's maintain consistency and use background here :)

a1tus:
Done, but there're a few cases to discuss.

1) I've added docs, but probably padding could be also mentioned in "About the processing pipeline".

2) Coz we need to use `background` option now implicit flatten could be applied. Default bg color is white, so we can get different results for calls without bg at all and with smth like `/bg:fff/`. Not sure if it's bad. Just to consider.

3) We need to decide what should happen if padding that is more than image size is set.

I've adding this one (not commited) to get rid of strange image transformations.

```
	if po.Padding.Enabled {
		dstW -= float64(po.Padding.Left + po.Padding.Right)
		dstH -= float64(po.Padding.Top + po.Padding.Bottom)
		if dstW < 1 {
			dstW = 1
		}
		if dstH < 1 {
			dstH = 1
		}
	}
```

But still in this case resulting image size grows coz of this part (image is resized to 1x1px + surrounded with passed padding):

```
		img.Embed(
			img.Width()+po.Padding.Left+po.Padding.Right,
			img.Height()+po.Padding.Top+po.Padding.Bottom,
			po.Padding.Left,
			po.Padding.Top,
			po.Background,
		)
```

We can make code more complex and prevent image growing or leave it for end users.

4) Related to previous. Extend goes after padding. Enlarge goes before. Some results can be not obvious. Not sure if we should cover in docs (I doubt that it's normal usage).

DarthSim:
> I've added docs, but probably padding could be also mentioned in "About the processing pipeline".

"About the processing pipeline" needs updating anyway, I'll do this

> Coz we need to use background option now implicit flatten could be applied. Default bg color is white, so we can get different results for calls without bg at all and with smth like /bg:fff/. Not sure if it's bad. Just to consider.

`vipsImage.Extend` works in a hacky way: if the image has alpha-channel, it extends it with fully transparent color, otherwise, it uses the provided color. In case you don't need to add a border with a color that differs from the background, it's ok.

> We need to decide what should happen if padding that is more than image size is set

Good catch. I wanted to point on this myself. I think, increasing `srcW` and `srcH` instead of decreasing `dstW` and `dstH` should solve the problem. IMHO if the user sets the padding larger than the requested size, they should expect some unexpected behavior :)

The other solution is making padding outbound and moving padding after extending. This would solve 4 too.

DarthSim:
I think outbound padding would be a better solution since it solves most of the problems and has a more predictable behavior.

a1tus:
I've tried to move it under extend like that:

	if po.Extend.Enabled && (po.Width > img.Width() || po.Height > img.Height()) {
		offX, offY := calcPosition(po.Width, po.Height, img.Width(), img.Height(), &po.Extend.Gravity, false)
		if err = img.Embed(po.Width, po.Height, offX, offY, po.Background); err != nil {
			return err
		}
	}

	if po.Padding.Enabled {
		if err = img.Embed(
			img.Width()+po.Padding.Left+po.Padding.Right,
			img.Height()+po.Padding.Top+po.Padding.Bottom,
			po.Padding.Left,
			po.Padding.Top,
			po.Background,
		); err != nil {
			return err
		}
	}

	checkTimeout(ctx)

	if po.Watermark.Enabled && watermark != nil {
		if err = applyWatermark(img, watermark, &po.Watermark, 1); err != nil {
			return err
		}
	}

	return img.RgbColourspace()

...and it works well. Also this allowed to remove padding calc in `calcScale`.

But now we have 2 new questions:

1) Do we need to apply dpr for padding?

2) May padding embed should go after watermark too?

DarthSim:
> Do we need to apply dpr for padding?

Damn, that's a good question. Seems like we should.

> May padding embed should go after watermark too?

I don't think so:
1. `extend` is applied before watermark, let's maintain consistency.
2. There is an optimization for applying watermarks on animated images that wouldn't work if watermarks are applied not at the last step.

a1tus:
To summarize:
1) We leave padding code exactly like in snippet above.
2) Do apply dpr for padding (in simple terms enlarge it).
3) Mention in docs that padding is applied after all operation except watermark and encounters dpr.

DarthSim:
Yup, that's right. Sorry for the delay

a1tus:
Now I'm sorry for the delay :) All that stuff happening made my working computer unavailable for some time.

I've updated PR, seems like now it's all like we discussed.

DarthSim:
Looks good! Thanks for your contribution!

PR #369, opened 2020-04-01
Generating secrets for Heroku automatically

bmann:
Your deploy to heroku wasn't using the `generator: secret`. I added that to app.json, and it now does deploy automatically with the secrets filled in.

### DIFF
--- a/app.json
+++ b/app.json
@@ -9,12 +9,12 @@
   "env": {
     "IMGPROXY_KEY": {
       "description": "Hex-encoded secret key for signing URLs.",
-      "value": "",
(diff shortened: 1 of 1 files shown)


DarthSim:
Hi,
`IMGPROXY_KEY` and `IMGPROXY_SALT` should be hex-encoded. The documentation says the following:

> Currently the only supported generator is secret, which generates a pseudo-random string of characters.

A pseudo-random string would most probably not be a hex-encoding compatible one.

bmann:
@DarthSim I installed it and it worked using this 🤷

Issue #220, opened 2019-10-16
Fallback image (or: default / placeholder / 404 image)


aight8:
When talk about application side image processing, default images are a pain too. The implementation is very fragmented and requires lot of extra logic, placeholder right size.
It could nice together with cachebuster.

The implementation could be similar to watermark. Defining one fallback image would be nice, define it per preset or query even more nicer.

At best, the fallback image can have a full processing pipeline. With croping/resizing/fitting etc. it can create out of a prototype fallback image one, which can be used everywhere. (imagine an icon with a simple background)

Evtl. it makes sense to set the default image size of the fallback image same to the requested size of the original one (if exists). Just an idea.

In term of a proxy this is just defining the 404 page (or image here). So it would be totally legitim and keep imgproxy simple.

*Sidenote for devs*
If you have a deterministic source url for your entities primary image (like: /user/5343.png), you can automatically fetch the image or fallback image only with the users ID. In this case, Deterministic URL, requires an application side signing/encrypting for security.

DarthSim:
Implemented by @ehiggs-deliverect 👍

PR #374, opened 2020-04-07
Implement fallback images in a similar fashion to watermarks.

ehiggs-deliverect:
Addresses #220 where, if we cannot retrieve the specified file for the user, we use a fallback image which is then manipulated in a consistent way as the request.

Use the following configuration to set a fallback image which is cached at startup:

`IMGPROXY_FALLBACK_DATA` - a base64 encoded image to be used in case of a fallback.
`IMGPROXY_FALLBACK_FILE` - a local file to be used for the fallback image.
`IMGPROXY_FALLBACK_URL` - a URL to be used for the fallback image.

EDIT:
`IMGPROXY_FALLBACK_IMAGE_DATA` - a base64 encoded image to be used in case of a fallback.
`IMGPROXY_FALLBACK_IMAGE_FILE` - a local file to be used for the fallback image.
`IMGPROXY_FALLBACK_IMAGE_URL` - a URL to be used for the fallback image.


fallback_data.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, fmt.Errorf("Can't decode fallback data: %s", err)
+	}
+
+	imgtype, err := checkTypeAndDimensions(bytes.NewReader(data))
+	if err != nil {
+		return nil, fmt.Errorf("Can't decode fallback: %s", err)
+	}
+
+	return &imageData{Data: data, Type: imgtype}, nil
+}
```
DarthSim, line 38: Let's rename `base64WatermarkData` to `base64ImageData` and reuse it instead of duplicating the code. Same for the further functions. All the `imageData` loading functions and the type itself can be put to `image_data.go`.

processing_handler.go
```diff
@@ -186,3 +194,11 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
...
 	respondWithImage(ctx, reqID, r, rw, imageData)
 }
+
+func loadFallback() (err error) {
+	fallback, err = getFallbackData()
+	if err != nil {
+		logError("Could not load fallback data. Fallback images will not be available: %s", err.Error())
+	}
+	return
+}
```
DarthSim, line 204: `getFallbackData` error should not be suppressed. Suppressing the error would introduce an unpredictable behavior.

DarthSim:
Thanks for your contribution!

ehiggs-deliverect:
Awesome! Thanks for the quick merge. :)

PR #386, opened 2020-04-25
Basic support for PDF

peterlozano:
Libvips has support for PDF when PDFium or Poppler are availble.

I've tried this patch and imgproxy is able to handle pdfs.

However, I'm not sure how to expose to the user of Imgproxy the possibility of choosing a page from the pdf, or an scale or dpi.

It can be done by changing vips.c and passing the additional parameters documented [here](https://github.com/libvips/libvips/blob/df72de89d0835e701b2903e68bc369f4a6676bf7/libvips/foreign/pdfload.c#L902). 

```
  return vips_pdfload_buffer(buf, len, out, "access", VIPS_ACCESS_SEQUENTIAL, "page", 4, "scale", 4.0, "dpi", 400.0, NULL);
```

but, as I said, I'm not sure how this could be exposed to the user in the url.

Thanks for imgproxy!

### DIFF
--- a/image_type.go
+++ b/image_type.go
@@ -26,6 +26,7 @@ const (
 	imageTypeHEIC    = imageType(C.HEIC)
 	imageTypeBMP     = imageType(C.BMP)
 	imageTypeTIFF    = imageType(C.TIFF)
+	imageTypePDF     = imageType(C.PDF)
 
 	contentDispositionFilenameFallback = "image"
 )
@@ -42,6 +43,7 @@ var (
 		"heic": imageTypeHEIC,
 		"bmp":  imageTypeBMP,
 		"tiff": imageTypeTIFF,
+		"pdf":  imageTypePDF,
 	}
 
 	mimes = map[imageType]string{
@@ -54,6 +56,7 @@ var (
 		imageTypeHEIC: "image/heif",
 		imageTypeBMP:  "image/bmp",
 		imageTypeTIFF: "image/tiff",
+		imageTypePDF:  "application/pdf",
 	}
 
 	contentDispositionsFmt = map[imageType]string{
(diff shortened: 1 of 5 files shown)


DarthSim:
Hi Pedro,

I think PDF should be processed just like SVG - load it with a proper scale. No need to change DPI.

We have a problem with dependencies. Poppler is licensed under GPLv2. I don't like to use GPL libs in OSS version, and we can't them in Pro. Linking libvips with PDFium is a bit fancy, I need to try it to determine if I can build Docker image with it.

peterlozano:
I agree, I don't like Poppler either. I wanted to use PDFium but, as you said, it is super tricky to link libvips with it.

vicmosin:
Any updates?

DarthSim:
Sorry for the delay, I was on vacation. I'll try to return to this soon.

DarthSim:
Ok, I managed to build vips with PDFium support. Let's finish this PR.

There's what need to be done:
1) Remove PDF saving code since we not going to support PDF saving.
2) Do scale on load the same way we do it with SVG.

DarthSim:
Since there's no activity, proper implementation of PDF support goes right to Pro.

peterlozano:
Sorry @DarthSim, I was hoping a client would pay for my time to work on this but they didn't accept it, and I haven't found free time to finish it.

Thorleon:
@DarthSim I may be interested in finishing this MR, would you reconsider PDF support in public version if I solve this pull request?

DarthSim:
@Thorleon I have already released Pro version with PDF support, so it won't go to the OSS version. Anyway, feel free to fork this repo and build your own image.

vicmosin:
@DarthSim how can I get in touch with your team regarding PRO prices? No response from your online form at website

DarthSim:
@vicmosin sorry for the delay. Though our manager tries to respond asap, it's not always the case :( I'll bump your application.

Issue #421, opened 2020-06-21
Add keep meta data as a processing option


jkoehl:
We have cases where we want to strip meta data from the response images and other cases where we would like to override this on some requests.

I have a PR prepared which adds this capability to the processing options.

DarthSim:
The feature is released in `v2.14.0`

PR #422, opened 2020-06-21
Implements #421 processing option to control meta stripping

jkoehl:
Implements #421

### DIFF
--- a/docs/generating_the_url_advanced.md
+++ b/docs/generating_the_url_advanced.md
@@ -431,11 +431,11 @@ Default: empty
 #### Keep Meta
 
 ```
-keepmeta:%keepmeta
-km:%keepmeta
+stripmeta:%stripmeta
+sm:%stripmeta
 ```
 
-When set to `1`, `t` or `true`, imgproxy will keep the metadata (EXIF, IPTC, etc.) on JPEG and WebP output images. Normally this is controlled by the [IMGPROXY_STRIP_METADATA](configuration.md#miscellaneous) configuration setting but this procesing option allows the configuration to be overridden for a specific request.
+When set to `1`, `t` or `true`, imgproxy will strip the metadata (EXIF, IPTC, etc.) on JPEG and WebP output images. Normally this is controlled by the [IMGPROXY_STRIP_METADATA](configuration.md#miscellaneous) configuration but this procesing option allows the configuration to be set for each request.
 
 Default: `false`
 
--- a/docs/generating_the_url_advanced.md
+++ b/docs/generating_the_url_advanced.md
@@ -428,11 +428,11 @@ It's highly recommended to prefer `cachebuster` option over URL query string bec
 
 Default: empty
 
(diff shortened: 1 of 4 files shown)


processing_options.go
```diff
@@ -848,6 +850,20 @@ func applyFilenameOption(po *processingOptions, args []string) error {
 	return nil
 }
 
+func applyMetaStripOption(po *processingOptions, args []string) error {
+	nArgs := len(args)
+
+	if nArgs > 1 {
```
DarthSim, line 856: This can be squashed to `if len(args) > 1` since we don't use `nArgs` further

process.go
```diff
@@ -639,7 +639,7 @@ func saveImageToFitBytes(po *processingOptions, img *vipsImage) ([]byte, context
 	img.CopyMemory()
 
 	for {
-		result, cancel, err := img.Save(po.Format, quality, conf.StripMetadata)
+		result, cancel, err := img.Save(po.Format, quality, conf.StripMetadata && !po.KeepMeta)
```
DarthSim, line 642: `conf.StripMetadata && !po.KeepMeta` gives us a wrong result in case both `conf.StripMetadata` and `po.KeepMeta` are falsey. The best way to fix this is to set `po.KeepMeta` default value from config and use only `!po.KeepMeta` here.

processing_options.go
```diff
@@ -894,6 +910,8 @@ func applyProcessingOption(po *processingOptions, name string, args []string) er
 		return applyCacheBusterOption(po, args)
 	case "filename", "fn":
 		return applyFilenameOption(po, args)
+	case "keepmeta", "km":
```
DarthSim, line 913: Let's maintain consistency and name it `strip_metadata` (`sm`)

docs/generating_the_url_advanced.md
```diff
@@ -428,6 +428,17 @@ It's highly recommended to prefer `cachebuster` option over URL query string bec
 
 Default: empty
 
+#### Keep Meta
+
+```
+stripmeta:%stripmeta
```
DarthSim, line 434: Still, let's name it `strip_metadata`. The same goes for the `ProcessingOptions` field.

docs/generating_the_url_advanced.md
```diff
@@ -428,6 +428,17 @@ It's highly recommended to prefer `cachebuster` option over URL query string bec
 
 Default: empty
 
+#### Keep Meta
```
DarthSim, line 431: I believe this should be `Strip metadata` :)

processing_options.go
```diff
@@ -848,6 +850,14 @@ func applyFilenameOption(po *processingOptions, args []string) error {
 	return nil
 }
 
+func applyMetaStripOption(po *processingOptions, args []string) error {
```
DarthSim, line 853: `applyStripMetadataOption`

DarthSim:
Thanks for your contribution again!

PR #461, opened 2020-08-22
Debug headers

### LINKED ISSUE #452: Debug headers?
Hi, great project!

Wondering if would be possible to add some kind of debugging information in the response headers. Something along the lines of below.

```
x-imgproxy-bytes-before: 123456;
x-imgproxy-bytes-after: 12345;
```

deckchairlabs:
I've never written go code before, but thought I would have a "go" at implementing the Debug Headers feature. I think I'm doing it right, I used your darthsim/imgproxy-base Docker image for testing and building, all seems to work as it should.

Fixes #452

### DIFF
--- a/config.go
+++ b/config.go
@@ -295,7 +295,7 @@ var conf = config{
 	SentryEnvironment:              "production",
 	SentryRelease:                  fmt.Sprintf("imgproxy/%s", version),
 	ReportDownloadingErrors:        true,
-	EnableDebugHeaders:				false,
+	EnableDebugHeaders:             false,
 	FreeMemoryInterval:             10,
 	BufferPoolCalibrationThreshold: 1024,
 }
--- a/config.go
+++ b/config.go
@@ -263,6 +263,8 @@ type config struct {
 
 	ReportDownloadingErrors bool
 
+	EnableDebugHeaders bool
+
 	FreeMemoryInterval             int
 	DownloadBufferSize             int
 	GZipBufferSize                 int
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -293,6 +295,7 @@ var conf = config{
 	SentryEnvironment:              "production",
 	SentryRelease:                  fmt.Sprintf("imgproxy/%s", version),
 	ReportDownloadingErrors:        true,
+	EnableDebugHeaders:             false,
```
DarthSim, line 298: Boolean values are `false` by default. There's no need in this line.

processing_handler.go
```diff
@@ -107,6 +107,13 @@ func respondWithImage(ctx context.Context, reqID string, r *http.Request, rw htt
 		rw.Write(buf.Bytes())
 	} else {
 		rw.Header().Set("Content-Length", strconv.Itoa(len(data)))
+
+		if conf.EnableDebugHeaders {
+			imgdata := getImageData(ctx)
+			rw.Header().Set("X-Imgproxy-Bytes-Before", strconv.Itoa(len(imgdata.Data)))
+			rw.Header().Set("X-Imgproxy-Bytes-After", strconv.Itoa(len(data)))
+		}
```
DarthSim, line 115: I'd move this out of this `if`. Otherwise, this won't work for gzipped responses.

DarthSim (changes requested):
Hi James!

Thanks for your work! Please fix the things I pointed at and add the docs, and I'll merge this.

DarthSim:
Closing this PS due to #509. Thanks for the attempt anyway!

Issue #160, opened 2019-05-16
Possible incorrect handling of orientation (not sure if bad image/apple-special, or imgproxy)


mike-hart:
Some images, though they have an orientation EXIF tag set, do not get orientation corrected on resize.

I'm not sure if this is a problem with the image itself, or if there's something that needs looking into here.

The attached original image (click through, because weirdly the github resize renders right-way-up) renders upside down on Linux Firefox, Linux Slack, Linux Chrome, and Windows Chrome. Not sure about Apple.

![2728_property_2019_05_01_08_57_17_984134](https://user-images.githubusercontent.com/4129199/57836079-e0751200-77b7-11e9-83e1-2d548317ad8b.JPG)

DarthSim:
Hi,

It's the image problem. The image pixels are oriented correctly, but its `exif-ifd0-Orientation` header is set to 3, which means it's rotated 180 degrees. Every single application I've tried shows the image upside down.

The reason why Chrome shows one correctly right here is that it doesn't respect  `exif-ifd0-Orientation` header for `img` tags: https://stackoverflow.com/questions/42401203/chrome-image-exif-orienation-issue

mike-hart:
Thank you. That's... frustrating (since people who upload images with this particular issue would complain), but makes sense.

DarthSim:
Well, maybe it makes sense to add a config option to disable autorotation. If you have more images with incorrect `exif-ifd0-Orientation` than with correct ones, the option will help you a little.

mike-hart:
An interesting thought. I guess I'll need to go over the catalogue to find imges with an orientation set and judge which would affect more users.

Thanks.

PR #465, opened 2020-08-30
Config option to disable autorotation

mcjansen:
In issue #160 @DarthSim suggested:
_Well, maybe it makes sense to add a config option to disable autorotation. If you have more images with incorrect exif-ifd0-Orientation than with correct ones, the option will help you a little._

This PR implements the config option **auto_rotate**, which is **true** by default. 

I followed the existing config conventions. So you can specify it both via the environment variable **IMGPROXY_AUTO_ROTATE**, and via the **auto_rotate** (or **ar** shortcut) option in an advanced URL.\

It is useful in our setup because we, unfortunately, got a lot of images with incorrect EXIF orientation meta info.

### DIFF
--- a/processing_options.go
+++ b/processing_options.go
@@ -144,6 +144,7 @@ type processingOptions struct {
 	Blur          float32
 	Sharpen       float32
 	StripMetadata bool
+	AutoRotate    bool
 
 	CacheBuster string
 
@@ -229,6 +230,7 @@ func newProcessingOptions() *processingOptions {
 			Dpr:           1,
 			Watermark:     watermarkOptions{Opacity: 1, Replicate: false, Gravity: gravityOptions{Type: gravityCenter}},
 			StripMetadata: conf.StripMetadata,
+			AutoRotate:    conf.AutoRotate,
 		}
 	})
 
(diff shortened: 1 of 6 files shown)


DarthSim:
Hey Marco!

Thanks for your work! Things are not that easy though. 

1. There's `extractMeta` function that returns image dimensions and orientation. If the orientation tag says that the image is rotated by 90 degrees, it swaps width and height. Thus if you just disable rotation, you'll get a wrongly resized image.
2. imgproxy removes the orientation tag in the following cases: when it's configured to strip meta, when it rotates the image, and when the resulting format does not support meta. Thus if you just disable rotation, the orientation tag will or will not be saved depending on the configuration and the resulting format, which is not good.

As a workaround I'd recommend the following:

1. Add an additional argument to `extractMeta`, something like `considerOrientation`. If it's set to `false`, don't extract orientation and return right [here](https://github.com/imgproxy/imgproxy/blob/3882abae48ee23152d92bb05210ed3acdd6308be/process.go#L43). You can use `po.AutoRotate` as a value for this argument. Thus it will return zero angle and no flip for every image if auto-rotation is disabled.

2. Call `img.Rotate()` no matter what. libvips is smart enough to not do anything for zero angle.

processing_options.go
```diff
@@ -858,6 +860,14 @@ func applyStripMetadataOption(po *processingOptions, args []string) error {
 	return nil
 }
 
+func applyAutoRotateOption(po *processingOptions, args []string) error {
+	if len(args[0]) > 0 {
+		po.AutoRotate = parseBoolOption(args[0])
+	}
```
DarthSim, line 866: Please add proper validation as I did for `strip_metadata` https://github.com/imgproxy/imgproxy/commit/a1679e5e092236bb19a2dabbdf728f1e40a7f9d0

DarthSim:
Closing this PR as I added a proper implementation with https://github.com/imgproxy/imgproxy/commit/6da8885605240d68986d00c0fdd4fdffe22baa4b. Thanks for your work anyway!

PR #515, opened 2020-11-26
Fix C# sample

NicolasReyDotNet:
- Refactor C# sample
- Fix generation fail on some url supplied with queries parameters 
=> ie some supplied url like the following ended with a Forbiden error from imgproxy :
https://www.mywebsite.com/medias/myimg.png?context=rWHzdGVyfGltYWdlc3wxNTcwG5nfGltYWdlcy9oMGMvaGVkLzkzMzM5MzE2NzE1ODIucG5tfDlo6TM13DQ2OW0wMJg30W7kY0c2ZjZ4ZbgaZnNjxWErOpQ0NTk5MOI2MzFPZTY3MTgzYeYxYTgyMcBkZxQ4MmI&frz-v=49389

The sample is now more readable (as the java sample for example) and fix the above error

### DIFF
--- a/examples/signature.cs
+++ b/examples/signature.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.IO;
 using System.Security.Cryptography;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


DarthSim:
Thanks!

PR #518, opened 2020-12-01
Crop by aspect ratio

xdimedrolx:
We want to switch [ngx_http_imgproc](https://github.com/tommiv/ngx_http_imgproc) to imgproxy. Does imgproxy not support crop by aspect ratio? 

This PR implements crop by aspect ratio.


DarthSim:
Hey!
Do you have a use case where you need to change the aspect ratio without resizing the image?

xdimedrolx:
We normally use resize after cropping.
For example now when we receive a widescreen image from user we crop it 1:1 и 3:4 and then we can either resize it or not.
Now we will not be able to crop in this way without knowing the image size? Is it so?
[ImgIX](https://docs.imgix.com/apis/rendering/size/ar) has this function

DarthSim:
Sorry for the misunderstanding. I understand your technical use case, but I don't understand your product use case for that feature. It looks like resizing to fill is what you need.

PR #531, opened 2020-12-16
optionally exclude preset from signature

nayakravi:
@DarthSim, I saw your response to my issue where I requested if we could exclude presets from the signature. While I understand your reasoning, I think I did not fully elaborate on our use case. So, we work with images in the public domain, so we are not really concerned with the full resolution images being accessed. What we are concerned with is the ability to send arbitrary requests to our server and overloading it (or a DOS/DDOS attack). On the other hand, we do need to get multiple formats for the same image in the browser and having to construct a signed URL server side for every case is cumbersome. So, this PR is a very narrow one with a very limited scope. Let me know if this makes sense?

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,14 @@
+steps:
+- id: 'Build: Docker Image'
+  name: gcr.io/cloud-builders/docker
+  args:
+  - build
+  - -f
+  - docker/Dockerfile
+  - -t
+  - us.gcr.io/$PROJECT_ID/dashboard:$BRANCH_NAME$TAG_NAME
+  - .
+
+# Push docker images to container registry
+images:
+- us.gcr.io/$PROJECT_ID/dashboard:$BRANCH_NAME$TAG_NAME
--- a/config.go
+++ b/config.go
@@ -239,8 +239,10 @@ type config struct {
 
 	BaseURL string
 
-	Presets     presets
-	OnlyPresets bool
+	Presets                     presets
+	OnlyPresets                 bool
+	// ExcludePresetsFromSignature is only taken into account when OnlyPresets is true
+	ExcludePresetsFromSignature bool
 
 	WatermarkData    string
 	WatermarkPath    string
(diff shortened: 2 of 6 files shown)


DarthSim:
Sorry, I stand by my opinion: this feature looks insecure to me, and it adds a lot of confusion in the signing process. Try to add this feature documentation to https://github.com/imgproxy/imgproxy/blob/master/docs/signing_the_url.md, and you'll see it yourself.

> So, we work with images in the public domain, so we are not really concerned with the full resolution images being accessed.

Even if you don't pass any options, imgproxy will decode and encode the image. If the image is quite large and heavy to decode/encode like PNGs, that'll be a kind of attack on your server. The only way to properly protect yourself is the server-side signing of the whole path. And I don't want to add a feature that allows opening security breaches.

However, if you really need this feature, feel free to use your fork. That's the power of open source! :)

nayakravi:
I understand your sentiment. However, I wanted to understand this a bit
better:  "Even if you don't pass any options, imgproxy will decode and
encode the image. If the image is quite large and heavy to decode/encode
like PNGs, that'll be a kind of attack on your server. The only way to
properly protect yourself is the server-side signing of the whole path. And
I don't want to add a feature that allows opening security breaches"
Since we are signing the rest of the path, attackers cannot send arbitrary
image urls to imgproxy. They can only send authorized urls with pre-defined
presets. So, I'm not sure how they can successfully attack with large
images, etc? Unless you are suggesting that they could use a large
legitimate image url and simply call imgproxy with that signed url with
different presets? But how is that different from taking an image url that
serves a large image and simply replaying it over and over again? Also,
given that we use a CDN for caching responses, most of these replay attacks
will never make it to imgproxy anyways?

DarthSim:
If you exclude the presets part from the string you sign, the signature will be valid for the same URL without presets at all. Guess you have thousands of images, that are always resized and thus their original sized versions are never cached by your CDN. An attacker grabs the URLs from your site, removes the presets parts, and starts to flood your server. imgproxy has to encode all the images in their original sizes that you don't want it to do because this takes time.

Basically, excluding presets from the signed part is not much different from a combination of disabled signature, `IMGPROXY_ONLY_PRESETS`, and `IMGPROXY_BASE_URL`/`IMGPROXY_ALLOWED_SOURCES`. But when you disable the signature, you realize all the consequences. Excluding presets from the signature on the other hand gives you a false sense of security.

nayakravi:
So, the way I implemented this is that the preset can only be excluded from
the signature when IMGPROXY_ONLY_PRESETS is true. Otherwise, the full path
after the signature is expected to be signed. So, will imgproxy still allow
presets to be excluded? I may have missed that part..

PR #537, opened 2020-12-26
Initialize buffer_size_bytes metric by exponential buckets

alxrem:
As buffer_size_bytes measured in bytes, default buckets from .005 to 10 are practically unsuitable for images.

I propose to initialize buffer_size_bytes by buckets with exponential growth from 1K to 8M. This is most likely not the best solution, but looking better than the current one.


DarthSim:
Hey Alexey! This makes sense to me. Thanks for the contribution!

PR #585, opened 2021-03-17
Removes checking for mdat box before meta box.

bantonj:
I have some heif files that have mdat boxes before meta boxes that decode fine in other places, including libvips. I'm curious why that state throws an error? My (fairly limited) understanding of the container format is that the order of boxes is not important.


DarthSim:
I was pretty sure the order is important. Thanks for the fix!

PR #590, opened 2021-03-19
Skip processing option

### LINKED ISSUE #476: Ability to skip processing of some formats with Query params
This is related to https://github.com/imgproxy/imgproxy/issues/393 issue, but would be nice to have ability to skip image processing in some specific areas. For example: skip GIF processing in messages, but convert GIF to JPG for Profile image. 

So I think would be nice to have param (maybe skip_processing=true) in URL to solve this problem, not only in ENV variable.

skryukov:
Adds skip processing option

```
skip_processing:%extension1:%extension2:...:%extensionN
skp:%extension1:%extension2:...:%extensionN
```

When set, imgproxy will skip the processing of listed formats. Also available as `IMGPROXY_SKIP_PROCESSING_FORMATS` configuration.

**📝Note:** Processing can be skipped only when the requested format is the same as the source format.

**📝Note:** Video thumbnail processing can't be skipped.

Default: empty

Closes #476

### DIFF
--- a/processing_handler.go
+++ b/processing_handler.go
@@ -185,11 +185,11 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 	checkTimeout(ctx)
 
 	po := getProcessingOptions(ctx)
-	if len(conf.SkipProcessingFormats) > 0 || len(po.SkipProcessingFormats) > 0 {
+	if len(po.SkipProcessingFormats) > 0 {
 		imgdata := getImageData(ctx)
 
 		if imgdata.Type == po.Format || po.Format == imageTypeUnknown {
-			for _, f := range append(conf.SkipProcessingFormats, po.SkipProcessingFormats...) {
+			for _, f := range po.SkipProcessingFormats {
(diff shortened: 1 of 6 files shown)


processing_handler.go
```diff
@@ -184,12 +184,12 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
...
 
-	if len(conf.SkipProcessingFormats) > 0 {
+	po := getProcessingOptions(ctx)
+	if len(conf.SkipProcessingFormats) > 0 || len(po.SkipProcessingFormats) > 0 {
 		imgdata := getImageData(ctx)
-		po := getProcessingOptions(ctx)
 
 		if imgdata.Type == po.Format || po.Format == imageTypeUnknown {
-			for _, f := range conf.SkipProcessingFormats {
+			for _, f := range append(conf.SkipProcessingFormats, po.SkipProcessingFormats...) {
```
skryukov, line 192: I'm not sure, is `append` is good enough despite possible duplication? @DarthSim

Issue #541, opened 2021-01-22
DisableCompression turned on breaks if content-encoding: gzip is set.


ehiggs:
Some third party servers might (for questionable reasons, tbh) use `content-encoding: gzip`. As we use `Transport { ... DisableCompression: true...}` this results in a gzip encoded blob of data. When we attempt to extract metadata from this, libvips fails. 

I can reproduce this using `wget` which, if you don't specify `--compression=auto` will not decompress the downloaded data. Passing the resulting file to `vips thumbnail image.jpeg 120 thumb.jpeg` fails as well while specifying `--compression=auto` results in everything working.

ehiggs:
Relevant: https://superuser.blog/golang-http-gzip-compression

ehiggs:
fixed!

PR #595, opened 2021-03-22
Allow compression at the transport level.

ehiggs:
Solves #541

### DIFF
--- a/download.go
+++ b/download.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/imgproxy/imgproxy/v2/imagemeta"
+	"github.com/klauspost/compress/gzip"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


DarthSim:
LGTM, but there are a few things I'd like to change:
- I don't think that rewriting the response body is a good idea.
- We don't know the actual content length as the content is gzipped. We should pass zero content length in the case of gzip.

Let's change it a little bit like this:

```go
body := res.Body
contentLength := int(res.ContentLength)

if res.Header.Get("Content-Encoding") == "gzip" {
  // ...
  body = gzipBody
  contentLength = 0
}

imgdata, err := readAndCheckImage(body, contentLength)
```

download.go
```diff
@@ -200,6 +201,15 @@ func downloadImage(ctx context.Context) (context.Context, context.CancelFunc, er
 		return ctx, func() {}, err
 	}
 
+	if res.Header.Get("Content-Encoding") == "gzip" {
+		gzipBody, errGzip := gzip.NewReader(res.Body)
+		if errGzip != nil {
+			return ctx, func() {}, err
```
DarthSim, line 207:
```suggestion
			return ctx, func() {}, errGzip
```

ehiggs: Good catch. I changed the name of the err variable to satisfy the linter.

DarthSim:
LGTM now 👍 Thanks for the contribution!

PR #596, opened 2021-03-22
Datadog support

skryukov:
Adds Datadog tracing support.

Closes #371

PR #597, opened 2021-03-23
Feature/datadog

kai-modrzok:
I add initial Datadog support to imgproxy (as described in https://docs.datadoghq.com/tracing/setup_overview/setup/go). I used the same style as the NewRelic implementation used. To activate Datadog tracing the ENV Variable `IMGPROXY_DATADOG_SERVICE_NAME` needs to be set.


DarthSim:
Implemented in #596.
BTW, you're trying to use `po` before it's parsed.

PR #605, opened 2021-04-04
Add ENV setting for AVIF speed parameter

perosb:
Add `IMGPROXY_AVIF_SPEED` env setting for AVIF speed parameters.
https://github.com/libvips/libvips/blob/master/libvips/foreign/heifsave.c#L725

* speed: CPU effort, 0 slowest - 8 fastest, AV1 compression only
Use speed to control the CPU effort spent improving compression.
This is currently only applicable to AV1 encoders, defaults to 5.

### DIFF
--- a/config.go
+++ b/config.go
@@ -225,6 +225,7 @@ type config struct {
 	PngInterlaced         bool
 	PngQuantize           bool
 	PngQuantizationColors int
+	AvifSpeed             int
 	Quality               int
 	FormatQuality         map[imageType]int
 	GZipCompression       int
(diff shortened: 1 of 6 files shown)


perosb:
@DarthSim any chance of getting this merged?

IMHO avif is pretty useless with current speed. Being able to adjust makes it actually usable.
Especially with aom=3 which is much faster then aom=2 and rav1e.

DarthSim:
Hey @perosb!

Support of `speed` introduced in vips 8.10.2, and we need to check the version to keep backward compatibility. Please check out how I use `VIPS_SUPPORT_*` flags and do the same for `speed`.

DarthSim:
Thanks for your contribution!

PR #634, opened 2021-05-20
Add Github action to notify imgproxy-helm about updates

nepalez:
The action sends event to the [imgproxy-helm](https://github.com:imgproxy/imgproxy-helm) abount any update in the master branch of the main repository.

This [PR](https://github.com/imgproxy/imgproxy-helm/pull/25) adds another action handing the event and publishing an issue about the update. That's how we can shorten the cycle of updating Helm chart for the new version of the app.

---

@DarthSim For this action to work you should create a token with write access and register it in this repository as the `HELM_REPO_ACCESS_TOKEN`. Feel free to call me on Slack in case of any difficulties with this ))

### DIFF
--- a/.github/actions/notify-imgproxy-helm.yml
+++ b/.github/actions/notify-imgproxy-helm.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Repository Dispatch
         uses: peter-evans/repository-dispatch@v1.1.3
         with:
-          token: ${{ secrets.REPO_ACCESS_TOKEN }}
+          token: ${{ secrets.HELM_REPO_ACCESS_TOKEN }}
           repository: imgproxy/imgproxy-helm
           event-type: imgproxy-config-updated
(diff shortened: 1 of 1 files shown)


.github/actions/notify-imgproxy-helm.yml
```diff
@@ -0,0 +1,19 @@
...
+    paths:
+      - docs/configuration.md
+jobs:
+  notify-imgproxy-helm:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v1.1.3
+        with:
+          token: ${{ secrets.REPO_ACCESS_TOKEN }}
```
DarthSim, line 16: Let's change the name as the current one is quite ambiguous

```suggestion
          token: ${{ secrets.HELM_REPO_ACCESS_TOKEN }}
```

PR #644, opened 2021-06-01
Adds handling of negative height value for bmp files.

bantonj:
Windows bitmaps can have a negative height value, which can then be incorrectly decoded when using an unsigned int. This can make the megapixel calculation way too big.

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wmf/567172fa-b8a2-4d79-86a2-5e21d6659ef3

### DIFF
--- a/imagemeta/bmp.go
+++ b/imagemeta/bmp.go
@@ -33,7 +33,7 @@ func DecodeBmpMeta(r io.Reader) (Meta, error) {
 	} else {
 		// CORE
 		width = int(binary.LittleEndian.Uint16(tmp[18:20]))
-		height = int(binary.LittleEndian.Uint16(tmp[20:22]))
+		height = int(int16(binary.LittleEndian.Uint16(tmp[20:22])))
 	}
 
 	// height can be negative in Windows bitmaps
--- a/imagemeta/bmp.go
+++ b/imagemeta/bmp.go
@@ -29,7 +29,13 @@ func DecodeBmpMeta(r io.Reader) (Meta, error) {
 
 	if infoSize >= 40 {
 		width = int(binary.LittleEndian.Uint32(tmp[18:22]))
-		height = int(binary.LittleEndian.Uint32(tmp[22:26]))
+		signedHeight := int32(binary.LittleEndian.Uint32(tmp[22:26]))
+		// height can be negative in Windows bitmaps
+		if signedHeight < 0 {
+			height = int(-signedHeight)
+		} else {
+			height = int(signedHeight)
+		}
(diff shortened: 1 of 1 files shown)


imagemeta/bmp.go
```diff
@@ -29,7 +29,13 @@ func DecodeBmpMeta(r io.Reader) (Meta, error) {
...
 	if infoSize >= 40 {
 		width = int(binary.LittleEndian.Uint32(tmp[18:22]))
-		height = int(binary.LittleEndian.Uint32(tmp[22:26]))
+		signedHeight := int32(binary.LittleEndian.Uint32(tmp[22:26]))
+		// height can be negative in Windows bitmaps
+		if signedHeight < 0 {
+			height = int(-signedHeight)
+		} else {
+			height = int(signedHeight)
+		}
```
DarthSim, line 38: Here's a bit shorter version:

```suggestion
		height := int(int32(binary.LittleEndian.Uint32(tmp[22:26])))
		// height can be negative in Windows bitmaps
		if height < 0 {
			height = int(-height)
		}
```

Also, it seems like the same trick should be applied to 16-bit height. And we can move the negative height check out of this `if`.

DarthSim (changes requested):

DarthSim:
Hey!
Sorry for the delay, I was on vacation without access to GH.

Thanks for catching this! Please, fix the things I mentioned, and I'll merge this.

bantonj:
Vacation sounds nice, :)

I changed to your simplified version, and moved it down to include checking the 16-bit OS/2 version of the height. I'm pretty sure that it's supposed to be unsigned, but it doesn't hurt to include it in the check.

imagemeta/bmp.go
```diff
@@ -29,13 +29,18 @@ func DecodeBmpMeta(r io.Reader) (Meta, error) {
 
 	if infoSize >= 40 {
 		width = int(binary.LittleEndian.Uint32(tmp[18:22]))
-		height = int(binary.LittleEndian.Uint32(tmp[22:26]))
+		height = int(int32(binary.LittleEndian.Uint32(tmp[22:26])))
 	} else {
 		// CORE
 		width = int(binary.LittleEndian.Uint16(tmp[18:20]))
 		height = int(binary.LittleEndian.Uint16(tmp[20:22]))
```
DarthSim, line 36: Please add the same conversion (`int(int16(...))`) here.

DarthSim (changes requested):

DarthSim:
Thanks for your contribution!

PR #677, opened 2021-07-29
Support wildcards with * in AllowedSources (excluding slashes)

hlubek:
For some cases like CDNs or other use-cases where subdomains are involved there is no way to define `IMGPROXY_ALLOWED_SOURCES` with all possible URLs.

This change adds support with a wildcard character `*` that matches every non-slash character. It keeps matching of prefixes and uses Go regexps internally.

### DIFF
--- a/config.go
+++ b/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


DarthSim:
Hi Christopher!

The feature looks good, but I'd build and compile the regexp when parsing config. It's better to add a function like `patternEnvConfig` for this.

hlubek:
> Hi Christopher!
> 
> The feature looks good, but I'd build and compile the regexp when parsing config. It's better to add a function like `patternEnvConfig` for this.

Thanks for the feedback. Sure, it's better to do it at initialization of course. I'll update the PR.

DarthSim:
LGTM! Please, fix the linting errors and I'll merge this.

DarthSim:
Thanks for your contribution!

Issue #719, opened 2021-10-19
Pass along cookies


doerwalter:
We have imgproxy and the delivery service for the original images running on the same hostname (via several docker containers), so for example the original images would be `https://www.example.org/images/foo.png` and the URL for the resized image would be `https://www.example.org/imgproxy/insecure/rt:fill/w:42/h:42/plain/https://www.example.org/images/foo.png`.

Some of the images are login-protected, so they will only be delivered by the image delivery service if it gets the appropriate cookies. Those cookies all have the attribute `Path=/`, so they would be sent in requests to both URLs `https://www.example.org/images/foo.png` and `https://www.example.org/imgproxy/rt:fill/w:42/h:42/plain/https://www.example.org/images/foo.png`.

However imgproxy does not pass the cookie it gets in its request along in the request to the original URL although the destination URL matches the cookie domain and path.

The result is that the request to the original image will fail, so imgproxy can't be used to resize login-protected images.

It would be useful if imgproxy could be configurated to pass through cookies that match the destination URL.

For example the feature could be activated via the environment variable `IMGPROXY_COOKIE_PASSTHROUGH=true` and imgproxy would have to be able to determine on which hostname and path it is running. This could either be done via inspecting the `Host` or `X-Forwarded-Host` or by configuring it via additional environment variables.

PR #728, opened 2021-11-05
Implement cookie passthrough

doerwalter:
This pull request implements the feature requested by https://github.com/imgproxy/imgproxy/issues/719

It adds options to pass through incoming cookies in the image request if they are applicable.

This behaviour gets activated when the environment variable `IMGPROXY_COOKIE_PASSTHROUGH` is set (without it no cookies will be passed through).

From `docs/configuration.md`:

> When cookie forwarding is activated, imgproxy by default assumes the scope of the cookies to be all URLs with the same hostname/port and request scheme as given by the headers `X-Forwarded-Host`, `X-Forwarded-Port`, `X-Forwarded-Scheme` or `Host`. To change that use `IMGPROXY_COOKIE_BASE_URL`.
>
> * `IMGPROXY_COOKIE_PASSTHROUGH`: when `true`, incoming cookies will be passed through to the image request if they are applicable for the image URL. Default: false;
>
> * `IMGPROXY_COOKIE_BASE_URL`: when set, assume that cookies have a scope of this URL for the incoming request (instead of using the request headers). If the cookies are applicable to the image URL too, they will be passed along in the image request.

### DIFF
--- a/download.go
+++ b/download.go
@@ -10,7 +10,10 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/http/cookiejar"
+	"net/url"
 	"time"
+	"golang.org/x/net/publicsuffix"
 
 	"github.com/imgproxy/imgproxy/v2/imagemeta"
 )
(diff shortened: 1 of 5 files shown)


download.go
```diff
@@ -171,12 +174,55 @@ func readAndCheckImage(r io.Reader, contentLength int) (*imageData, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		cookieBase := conf.CookieBaseURL
+
+		if len(cookieBase) == 0 {
+			scheme := r.Header.Get("X-Forwarded-Proto")
+			if len(scheme) == 0 {
+				scheme = "https"
```
DarthSim, line 194: Since imgproxy doesn't support HTTPS, it seems that the default scheme should be `http`

DarthSim (changes requested):
Hey Walter!

The feature itself looks good, however, I'd like you to make some changes in the code:

1. The PR should be based on the `version/3` branch as the master is feature-frozen.

2. I don't like the idea of passing the request to `downloadImage`. I think there should be a sub-package that prepares the cookie jar, and that jar should be passed to `downloadImage` instead of request.

Please make these changes and I'll merge the PR.

PR #730, opened 2021-11-09
Implement cookie passthrough

doerwalter:
(This is an updated version of the pull request https://github.com/imgproxy/imgproxy/pull/728. Since I've recreated the branch, Github chose to close the pull request and I can't reopen it.)

This pull request implements the feature requested by #719

It adds options to pass through incoming cookies in the image request if they are applicable.

This behaviour gets activated when the environment variable `IMGPROXY_COOKIE_PASSTHROUGH` is set (without it no cookies will be passed through).

From `docs/configuration.md`:

>   When cookie forwarding is activated, imgproxy by default assumes the scope of the cookies to be all URLs with the same hostname/port and request scheme as given by the headers `X-Forwarded-Host`, `X-Forwarded-Port`, `X-Forwarded-Scheme` or `Host`. To change that use `IMGPROXY_COOKIE_BASE_URL`.
>
>   *   `IMGPROXY_COOKIE_PASSTHROUGH`: when true, incoming cookies will be passed through to the image request if they are applicable for the image URL. Default: false;
>
>   *   `IMGPROXY_COOKIE_BASE_URL`: when set, assume that cookies have a scope of this URL for the incoming request (instead of using the request headers). If the cookies are applicable to the image URL too, they will be passed along in the image request.

I've updated the pull request as requested:

* It's based on the branch `version/3`.
* The cookies are extracted via the new function `cookieJarFromRequest()` and only the `cookiejar.Jar` is passed to `downloadImage()` and other functions.
* I've updated the default scheme to `http`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -27,6 +27,7 @@ var (
 
 	TTL                     int
 	CacheControlPassthrough bool
+	CookiePassthrough       bool
 	SetCanonicalHeader      bool
 
 	SoReuseport bool
@@ -87,7 +88,8 @@ var (
 
 	ETagEnabled bool
 
-	BaseURL string
+	BaseURL       string
+	CookieBaseURL string
 
 	Presets     []string
 	OnlyPresets bool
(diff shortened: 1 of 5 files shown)


DarthSim:
Merged and polished a little bit. Thanks for your contribution!

PR #742, opened 2021-11-24
Use advanced url format for local serving example

bayandin:
The current example doesn't work with version 3.0


DarthSim:
Thank you! Missed this

PR #760, opened 2021-12-18
AvifSpeed value checker

AleksandrKrokis:
Hello, 

As I understand valid values for `AvifSpeed` is 0-8, including 0 and 8. But the checker does not allow 0.
So here is fix


DarthSim:
Thanks!

PR #765, opened 2021-12-22
fix: make text selection readable

Kiryous:
current setup results in low contrast between default text selection background color and text color. to fix this use link-color for text ::selection

## Before
Screenshot 2021-12-22 at 07 38 08

## After
Screenshot 2021-12-22 at 07 46 48


DarthSim:
Hey @Kiryous!
This makes sense, but your commit also adds background to headers. Please fix this and I'll merge the PR.

DarthSim:
Fixed it myself

PR #793, opened 2022-02-11
Add support for TextFormatter aka logfmt

perosb:
Add support for TextFormatter which is compatible with logfmt

image

This is very similar to the custom "structured" format.


perosb:
Our need for logfmt went away as we moved to json instead. Unless there is a need, this can be closed.

DarthSim:
Hey @perosb!

Sorry for the delay.

Since the requested log format needs to be polished a bit and you don't need it anymore, I'm going to close this PR. Thanks for the effort anyway!

PR #817, opened 2022-03-09
add config to fallback ttl to avoid large ttl when fallback responses

dcanob:


DarthSim:
Hey @dcanob!

Thanks for the effort, but I can't merge this PR as is. Basically, the param you added doesn't control TTL based on whether a fallback image is used or not. Instead, it controls TTL based on the response status code. The default status code for requests that use a fallback image is 200, so the param won't work by default. Also, we may add some cases where the status code will be changed, and then the parameter will work incorrectly.

dcanob:
thanks @DarthSim you are absolutely right I had not seen it that way, I generated a new PR this time using the FallbackImage and its Headers, I hope it helps.

Otherwise I would have to use a proxy to change the ttl of the images that imgproxy cannot process

PR #886, opened 2022-05-31
read image from form data

### LINKED ISSUE #852: Are there any plans to receive images directly?
Are there any plans to receive images directly?
eg: http `multipart/form-data` or `application/octet-stream`, instead of having to go through a URL or other OSS

xuzhijvn:
fixed #852

### DIFF
--- a/imagedata/image_data.go
+++ b/imagedata/image_data.go
@@ -1,9 +1,11 @@
 package imagedata
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/cookiejar"
 	"os"
@@ -142,4 +144,22 @@ func Download(imageURL, desc string, header http.Header, jar *cookiejar.Jar) (*I
 	}
 
 	return imgdata, nil
+}
+
+func FromFormData(r *http.Request, desc string) (*ImageData, error) {
+	image, header, err := r.FormFile("image")
+	defer image.Close()
+	if err != nil {
+		return nil, err
+	}
+	buf := bytes.NewBuffer(nil)
+	if _, err := io.Copy(buf, image); err != nil {
+		return nil, err
+	}
+	filenameParts := strings.Split(header.Filename, ".")
+
+	return &ImageData{
+		Data: buf.Bytes(),
+		Type: imagetype.Types[filenameParts[1]],
+	}, nil
 }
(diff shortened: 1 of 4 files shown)


DarthSim:
Hi @xuzhijvn! Could you explain a use case for this feature? imgproxy is designed to work like a reverse proxy so it doesn't look like POST support fits imgproxy well.

xuzhijvn:
> Hi @xuzhijvn! Could you explain a use case for this feature? imgproxy is designed to work like a reverse proxy so it doesn't look like POST support fits imgproxy well.

Hi @DarthSim! I think imgproxy is more like an image factory, give it a raw image and it will output a destination image.So the POST request is used to give it a binary image stream and it outputs a binary stream as well.

Java code maybe like this:

```
    public static void main( String[] args ) throws IOException {
        CloseableHttpClient httpClient = HttpClients.createDefault();
        HttpPost request = new HttpPost("http://localhost:8080/insecure/rs:fit:150:100:no:0/plain/stream@png");
        MultipartEntityBuilder builder = MultipartEntityBuilder.create();
        builder.addTextBody("field1", "yes", ContentType.TEXT_PLAIN);
        // This attaches the file to the POST:
        File f = new File("/Users/tony/Downloads/dog.jpeg");
        builder.addBinaryBody(
                "image",
                Files.newInputStream(f.toPath()),
                ContentType.APPLICATION_OCTET_STREAM,
                f.getName()
        );
        HttpEntity multipart = builder.build();
        request.setEntity(multipart);
        CloseableHttpResponse response = httpClient.execute(request);
        HttpEntity responseEntity = response.getEntity();
        System.out.println(responseEntity);

        FileUtil.writeFromStream(responseEntity.getContent(), "/Users/tony/Desktop/dogg.png");
    }
```

DarthSim:
As I said earlier, imgproxy is designed to work as a reverse proxy and it is not meant to be used this way. For your case, I'd recommend using the [vipsthumbnail](https://www.libvips.org/API/current/Using-vipsthumbnail.md.html) tool from the libvips distribution.

xuzhijvn:
> As I said earlier, imgproxy is designed to work as a reverse proxy and it is not meant to be used this way. For your case, I'd recommend using the [vipsthumbnail](https://www.libvips.org/API/current/Using-vipsthumbnail.md.html) tool from the libvips distribution.

ok, you are right

xuzhijvn:
> As I said earlier, imgproxy is designed to work as a reverse proxy and it is not meant to be used this way. For your case, I'd recommend using the [vipsthumbnail](https://www.libvips.org/API/current/Using-vipsthumbnail.md.html) tool from the libvips distribution.

ok, you are right

PR #902, opened 2022-06-23
enhancement: optimize performance

xiaofeige:
In my opinion, there is no need to wait for the IO complete. once the image caculation is done, the concurrency control strategy should let the request just go though, thus,  making the CPU fully used.

### DIFF
--- a/processing_handler.go
+++ b/processing_handler.go
@@ -202,17 +202,6 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// The heavy part start here, so we need to restrict concurrency
-	select {
-	case processingSem <- struct{}{}:
-	case <-ctx.Done():
-		// We don't actually need to check timeout here,
-		// but it's an easy way to check if this is an actual timeout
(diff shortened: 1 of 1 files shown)


xiaofeige:
> Hi @xiaofeige!
> 
> I thought about something like this myself but decided not to do so and there are several reasons for this:
> 
> 1. Having the source image downloading process out of concurrency control may lead to memory bloat. If we have more requests than allowed by concurrency, imgproxy will download all the source images from that requests. Thus client can't control how many source images can be stored in memory at a time and can easily get OOM.
> 2. Having the response writing process out of concurrency control doesn't make much sense. The response writing takes nanoseconds and isn't slowed down by slow clients.

yeah, you are right, but I think it's a trade-off. In my presure test, I noticed that there are much more memory left when CPU is 100% used. When I move the concurrency control to where I suggest,  memory didn't grow that fast, basicly the same as before., but avg qps improved a lot.

xiaofeige:
by the way, format & resize operation performance are very pool, do you known how to improve it by adjust the configuration?  thanks a lot. ^_^

PR #903, opened 2022-06-24
Respond with 404 when the source image can not be found in OpenStack Object Storage

joec4i:
Does what the title says so imgproxy doesn't throw a 500 when it should return 404.

Review please. 🙏

### DIFF
--- a/transport/swift/swift.go
+++ b/transport/swift/swift.go
@@ -2,7 +2,9 @@ package swift
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


transport/swift/swift.go
```diff
@@ -46,12 +47,25 @@ func (t transport) RoundTrip(req *http.Request) (resp *http.Response, err error)
...
+
 	if err != nil {
+		if errors.Is(err, swift.ObjectNotFound) || errors.Is(err, swift.ContainerNotFound) {
+			return &http.Response{
+				StatusCode: http.StatusNotFound,
+				Proto:      "HTTP/1.0",
+				ProtoMajor: 1,
+				ProtoMinor: 0,
+				Header:     header,
+				Body:       nil,
```
DarthSim, line 60: I'd add a body with the error message:

```suggestion
				Body:       io.NopCloser(strings.NewReader(err.Error())),
```

DarthSim (changes requested):
Hi @joec4i!

Everything looks good! Let's add a response body and I'll merge this.

transport/swift/swift.go
joec4i: Thanks for the quick review. It's updated now. Thanks @DarthSim

DarthSim:
Thanks for your contribution!

PR #919, opened 2022-07-12
Changed Header sizes in README

ImOwly:
Changed header size of Simplicity, Speed, and Security headers of README since the headers are a bit small and seem more like text than headers themselves, increasing header sizes distinguishes them from text and increases readability.

changes made:

Simplicity(Heading level 4) -> Simplicity(Heading level 3)

Speed(Heading level 4) -> Speed(Heading Level 3)

Security(Heading level 4) -> Security(Heading Level 3)


DarthSim:
Hey @ImOwly!
Thanks for the suggestions, but README is formatted as intended

PR #953, opened 2022-08-17
use UTF8 to encode url to base64

heysaad:
the example looks wrong to use UTF8 at one place and ASCII the other place.


DarthSim:
Thanks!

PR #969, opened 2022-09-06
Add default value to IMGPROXY_MAX_REDIRECTS documentation and clarify usage of value 0

tuomasva:
Default value taken from https://github.com/imgproxy/imgproxy/blob/ff3d3cedc4abbbc815472fa5a9171f61650ff881/config/config.go#L206

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -73,7 +73,7 @@ To check if the source image is SVG, imgproxy reads some amount of bytes; by def
 
 Requests to some image sources may go through too many redirects or enter an infinite loop. You can limit the number of allowed redirects:
 
-* `IMGPROXY_MAX_REDIRECTS`: the max number of redirects imgproxy can follow while requesting the source image. Default: `10`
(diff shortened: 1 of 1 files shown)


tuomasva:
I could not confirm it by looking at the code but the documentation should mention that `0` either mean 0 (no redirects) or  infinite, whichever is the case here.

docs/configuration.md
```diff
@@ -73,7 +73,7 @@ To check if the source image is SVG, imgproxy reads some amount of bytes; by def
 
 Requests to some image sources may go through too many redirects or enter an infinite loop. You can limit the number of allowed redirects:
 
-* `IMGPROXY_MAX_REDIRECTS`: the max number of redirects imgproxy can follow while requesting the source image
+* `IMGPROXY_MAX_REDIRECTS`: the max number of redirects imgproxy can follow while requesting the source image. When set to `0`, no redirecting is allowed. Default: `10`
```
DarthSim, line 76:
```suggestion
* `IMGPROXY_MAX_REDIRECTS`: the max number of redirects imgproxy can follow while requesting the source image. When set to `0`, no redirects are allowed. Default: `10`
```

DarthSim (changes requested):
Good point, thanks. Please change the phrase from the comment below, and I'll merge this.

PR #971, opened 2022-09-09
fix(docs): typo source url encryption

Rkallenkoot:
Could not find it in the sourcecode, since this is part of the pro version. I'm assuming the typo is just in the docs and not in source code. Someone with access to Pro should verify this


DarthSim:
Omg, thank you for noticing this before it's released :)

PR #995, opened 2022-10-05
OpenTelemetry support

DarthSim:


docs/configuration.md
```diff
@@ -395,6 +395,22 @@ imgproxy can send its metrics to Datadog:
 
 Check out the [Datadog](datadog.md) guide to learn more.
 
+## OpenTelemetry metrics
+
+imgproxy can send request traces to an OpenTelemetry collector:
+
+* `IMGPROXY_OPEN_TELEMETRY_ENDPOINT`: OpenTelemetry collector endpoint (`host:port`). Default: blank
```
Travis-Turner, line 402: Is "blank" supposed be null or some other empty value? essentially, my question—will this be clearly understood? I didn't recall seeing this in the docs before, but it's been a while

PR #1052, opened 2022-12-11
Implement Azure Identity Authentication for Azure Transport

xaevik:
**Resubmission of PR #1042 as its original changeset was lost**

## Summary

The purpose of this pull-request is to introduce support within the Azure Transport for more secure authentication vectors. Specifically, it allows the use of a [Service Principal](https://learn.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals) or [Managed Identity](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview) as opposed to using the [Access Key](https://learn.microsoft.com/en-us/azure/storage/common/storage-account-keys-manage?tabs=azure-portal).

There are inherit risks in utilizing the access key of a storage account, namely that it grants whoever has access to the key **full privileged rights** to that account. Microsoft generally recommends that you **do not use** this key for everyday access and that you should manage access through other means such as the ones noted above so that it can be properly controlled and audited through RBAC.

## Direct Dependency Changes

[+] github.com/Azure/azure-sdk-for-go/sdk/azcore
[+] github.com/Azure/azure-sdk-for-go/azidentity
[+] github.com/Azure/azure-sdk-for-go/azblob
[-] github.com/Azure/azure-storage-blob-go/azblob

## Authentication

Every effort has been made to ensure that the original authentication pathway was retained as to avoid creating a breaking change scenario. If the environment variable `IMGPROXY_ABS_KEY` is set, the transport will continue to authenticate via `NewSharedKeyCredential` and not attempt any other path.

For users that wish to use one of the newer authentication methods, they can simply unset the `IMGPROXY_ABS_KEY` variable and the transport attempt to authenticate via `NewDefaultAzureCredential`. For authentication via a Service Principal, the following environment variables must be set:

### Service principal with secret

* `AZURE_CLIENT_ID`: Specifies the client ID for your application registration
* `AZURE_TENANT_ID`: Specifies the tenant ID for your application registration
* `AZURE_CLIENT_SECRET`: Specifies the client secret for your application registration.

### Service principal with certificate
* `AZURE_CLIENT_ID`: Specifies the client ID for your application registration
* `AZURE_TENANT_ID`: Specifies the tenant ID for your application registration
* `AZURE_CLIENT_CERTIFICATE_PATH`: Specifies the path to a PFX or PEM-encoded certificate including private key
* `AZURE_CLIENT_CERTIFICATE_PASSWORD`: (optional) the password protecting the certificate file (PFX (PKCS12))
* `AZURE_CLIENT_CERTIFICATE_CHAIN`: (optional) send certificate chain in x5c header to support subject name / issuer-based authentication


transport/azure/azure.go
```diff
@@ -32,44 +34,65 @@ func New() (http.RoundTripper, error) {
...
 
-	if config.ETagEnabled && start == 0 && end == azblob.CountToEnd {
-		etag := string(get.ETag())
-
-		if etag == req.Header.Get("If-None-Match") {
-			if body := get.Response().Body; body != nil {
-				get.Response().Body.Close()
-			}
+	if config.ETagEnabled {
+		azETag := string(*result.ETag)
```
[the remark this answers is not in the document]
xaevik: Forgive my assumptions here as I'm relatively new to the language, but in what scenario would this actually happen? Unless I'm understanding something wrong it could never be `nil` because it is not a pointer so its default value should be false if not set.

Based on the usage of this particular variable in both `s3.go` and `gcs.go` there is only a simple conditional check to see if it is `true`.

xaevik:
> Hi @xaevik! Please add support for ranged requests and fix the things I commented

I'm experiencing a road block in this implementation. `req.Header.Get("Range")` is always nil even when passing the Range header via cURL. Any input on this would be great so that I can do appropriate end-to-end testing.

transport/azure/azure.go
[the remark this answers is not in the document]
DarthSim: `result.ETag` is a pointer to `string`, so it can be `nil`. If you try to unref `nil`, you'll get SEGFAULT. Example: https://go.dev/play/.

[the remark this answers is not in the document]
xaevik: I misinterpreted your original comment, my apologies. I'll push up the refactor shortly.

transport/azure/azure.go
```diff
@@ -32,51 +44,119 @@ func New() (http.RoundTripper, error) {
...
+	if r := req.Header.Get("Range"); len(r) != 0 {
+		start, end, err := httprange.Parse(r)
+		if err != nil {
+			return httprange.InvalidHTTPRangeResponse(req), err
+		}
+
+		if end != 0 {
+			length := end - start + 1
+			if end < 0 {
+				length = -1
```
DarthSim, line 88: While it looks like `-1` should work here, I'd use `blob.CountToEnd`

transport/azure/azure.go
```diff
@@ -32,51 +44,119 @@ func New() (http.RoundTripper, error) {
...
 				Request:       req,
-			}, nil
+			}, err
 		}
 	}
 
-	return get.Response(), nil
+	header.Set("Accept-Ranges", "bytes")
+	header.Set("Content-Length", strconv.FormatInt(*result.ContentLength, 10))
+	header.Set("Content-Type", *result.ContentType)
```
DarthSim, line 141: I'd check that `result.ContentLength` and `result.ContentType` aren't `nil`. it's better safe than sorry :)

DarthSim (changes requested):

transport/azure/azure.go
xaevik: Unfortunately, under the newer azblob package, that property no longer exists in `azblob`.

transport/azure/azure.go
```diff
@@ -32,44 +34,65 @@ func New() (http.RoundTripper, error) {
 		return nil, err
 	}
 
-	serviceURL := azblob.NewServiceURL(*endpointURL, pipeline)
+	if config.ABSKey != "" {
```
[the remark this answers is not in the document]
xaevik: The null guard for `ABSName` is further up the code on line 34 as it is referenced multiple times.

transport/azure/azure.go
xaevik: It appears that this constant has been moved to `blockblob`, I am making the appropriate refactor now.

transport/azure/azure.go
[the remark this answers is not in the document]
DarthSim: Ah, that makes sense. We still need it to build the endpoint, I missed that

transport/azure/azure.go
```diff
@@ -32,51 +45,125 @@ func New() (http.RoundTripper, error) {
...
+				StatusCode:    azError.StatusCode,
+				Proto:         "HTTP/1.0",
+				ProtoMajor:    1,
+				ProtoMinor:    0,
+				Header:        header,
+				ContentLength: int64(body.Len()),
+				Body:          io.NopCloser(body),
+				Close:         false,
+				Request:       req,
+			}, err
```
DarthSim, line 117: We shouldn't return `err` here since we're returning a response.

transport/azure/azure.go
```diff
@@ -32,51 +45,125 @@ func New() (http.RoundTripper, error) {
...
 				ProtoMajor:    1,
 				ProtoMinor:    0,
-				Header:        make(http.Header),
+				Header:        header,
 				ContentLength: 0,
 				Body:          nil,
 				Close:         false,
 				Request:       req,
-			}, nil
+			}, err
```
DarthSim, line 136: We shouldn't return `err` here since we're returning a response.

transport/azure/azure.go
```diff
@@ -32,51 +45,125 @@ func New() (http.RoundTripper, error) {
...
+		StatusCode:    statusCode,
+		Proto:         "HTTP/1.0",
+		ProtoMajor:    1,
+		ProtoMinor:    0,
+		Header:        header,
+		ContentLength: *result.ContentLength,
+		Body:          ctxreader.New(req.Context(), result.Body, true),
+		Close:         true,
+		Request:       req,
+	}, err
```
DarthSim, line 168: We shouldn't return `err` here

DarthSim (changes requested):
Sorry for my inattentiveness. Added a few more comments

DarthSim:
Looks good now 👍 Thanks for your contribution!

PR #1054, opened 2022-12-14
fix: rendering the pro image

0xflotus:
I fixed the rendering of the pro image in the title of some of the doc files.

See: 

pro


DarthSim:
Thanks!

PR #1061, opened 2022-12-23
Update jemalloc package install instructions

gaborszakacs:
as `libjemalloc1` is not available in recent distributions anymore, see: https://packages.debian.org/search?keywords=jemalloc


DarthSim:
Hey @gaborszakacs! You're right. Thanks!

PR #1068, opened 2023-01-01
year update

IcedMonk:
updated the year to the current year


DarthSim:
Hi @IcedMonk! thanks for the effort, but if a copyright notice contains only a single year, it should not be updated

Issue #1126, opened 2023-03-06
How do I use the prebuilt docker image with local filesystem?


rouilj:
I develop an issue tracking system. People add images to the issues (screenshots etc.). I would like to use imgproxy to optimize the delivery of these images.

The images are stored in files on local disk. My first thought was to replace an image URL that
normally looks like this:
```
    https://example.com/issues/file14/face.jpg 
```
with (URL wrapped for readability; I explain why I am using expires below):
```
    https://example.com/imgproxy/security_hash/rs:fit:300:200:0:0/
           filename:face/expires:XXXXX/plain/local:///0/file14 
```
Then I run imgproxy with `IMGPROXY_LOCAL_FILESYSTEM_ROOT=/images` to reference the data stored in the file
```
   ~tracker/db/files/0/file14
```
and mount `~tracker/db/files` on `/images` in the docker volume.

I tried this, but it does not look like the distributed docker image has a volume/mount point exposed
according to docker inspect.

The `expires` value `XXXX` will be set to  5 seconds after the web page request has been received. Using expires is an attempt to prevent the URL from being used by another party. Images have access controls
in the tracker that determine if the image is accessible to the user. So I use a short duration time to allow the user to access it while the web page is loading, but not be able to use the URL outside of the loading time for the current page. Granted the user can screenshot the image etc. however this
is additional protection to prevent direct access to imgproxy from leaking potentially sensitive data
since it can't authenticate/authorize the user.

So three questions:

  1) can the distributed docker image be generated with a volume at `/images` (or other suitable mountpoint)

  2) does my proposed deployment method sound workable

  3) will the use of `expires` help "secure" the image data
 
Thanks and have a great week.

Also I added: https://github.com/imgproxy/imgproxy/pull/1135 to add /images volume to the
Dockerfile and image.

[ edited replacement url to put / between the processing options components and fix the resize option argument, and remove explicitly requesting a file type to set imgproxy choose the best
alternative. Added pull request]

DarthSim:
Hi @rouilj!

If your images are located in the filesystem available to your Docker engine, then using the `local:///` scheme is the most optimal way.

If you run imgproxy with `docker run`, you can use the `-v` flag to mount you:

```
docker run -v /path/to/local/dir:/images -it darthsim/imgproxy:latest
```

If you run imgproxy with `docker service run` you can use the `--mount` flag:

```
[...]

PR #1135, opened 2023-03-06
Allow mounting an external directory for local:/// storage.

rouilj:
Reference https://github.com/imgproxy/imgproxy/issues/1126

This change allows imgproxy to be configured to serve image files from a local filesystem mounted on /images.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,6 +32,9 @@ COPY --from=0 /usr/local/lib /usr/local/lib
 
 COPY NOTICE /usr/local/share/doc/imgproxy/
 
+RUN mkdir /images
+VOLUME /images
+
 ENV VIPS_WARNING=0
 ENV MALLOC_ARENA_MAX=2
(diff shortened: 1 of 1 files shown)


DarthSim:
Hi @rouilj!
You don't need these changes to mount a local directory to a container. You can just use the `-v` flag of `docker run`:

```
docker run -v /path/to/local/dir:/images -it darthsim/imgproxy:latest
```
`docker service run` doesn't support the `-v` flag, but supports the `--mount` flag:

```
docker servive run --mount type=bind,src=$(pwd)/www,dst=/www -t darthsim/imgproxy:latest
```

rouilj:
Interesting. I still think it's a good affordance to have though.

As I mentioned on the ticket, I didn't realize that you could mount into arbitrary or non-existent locations
in a docker container.

DarthSim:
The `VOLUME` instruction will make a volume every time the image is run. I believe this is redundant for most users.

Issue #348, opened 2020-02-26
Support for conditional request pass-through


Medo42:
I really like the decision of imgproxy to focus on one concern and explicitly exclude caching from its list of goals. However, I think the current pass-through support for putting a caching proxy in front of imgproxy is incomplete and should be extended for better behavior and performance.

When a client (either a browser or a caching proxy) has a cached version of the resource that they believe to be stale, they may send a conditional request to imgproxy with an If-Modified-Since or If-None-Match header. Currently, imgproxy appears to ignore If-Modified-Since, and only supports its own ETag generation for If-None-Match. In either case, the resource is always loaded from the origin server with an unconditional request.
My understanding is that imgproxy applies a deterministic transformation to the source image which only depends on the original resource and the transformation parameters encoded in the URL. This means that, if the original resource does not change, neither does the output of imgproxy. So I think it would be valid for imgproxy to pass If-Modified-Since and If-None-Match headers in requests it receives on to the requests to the origin server. If the response from the origin server is a 304 Not Modified, imgproxy can in turn answer its client with a 304.
Making this work would also require imgproxy to copy Last-Modified and ETag-Headers from origin responses to its own responses, although care might have to be taken with the semantics of strong and weak ETags (imgproxy might want to reversibly rewrite ETags from the origin to turn them weak).

DarthSim:
That's a good idea. Thank you!

celadaguido:
+1

celadaguido:
you have to add it to the ctx in download.go and then add it to the response in processing_handler.go.

PR #1147, opened 2023-04-02
Support Last-Modified response header and support If-Modified-Since request header.

ehiggs:
Support responding with the Last-Modified of the upstream image. This is a very useful header to determine if we are dealing with a new or old image. As we allow returning the upstream ETag, it makes sense to also return the upstream Last-Modified header.

Also support the `If-Modified-Since` request header to help manage caching. previously discussed in #348.


processing_handler.go
```diff
@@ -264,6 +273,12 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if modifiedSince := r.Header.Get("If-Modified-Since"); len(modifiedSince) != 0 {
+		if _, ok := r.Header["If-None-Match"]; !ok {
```
DarthSim, line 277: We should send both `If-Modified-Since` and `If-None-Match`. We don't know if the server supports ETags, so we should let the server decide which header to use. After all, we add `If-Modified-Since` support for the cases where the source server doesn't support ETags

processing_handler.go
```diff
@@ -303,8 +318,10 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 
 	if err == nil {
 		defer originData.Close()
-	} else if nmErr, ok := err.(*imagedata.ErrorNotModified); ok && config.ETagEnabled {
-		rw.Header().Set("ETag", etagHandler.GenerateExpectedETag())
+	} else if nmErr, ok := err.(*imagedata.ErrorNotModified); ok {
+		if config.ETagEnabled {
```
DarthSim, line 322: We should also check `len(etagHandler.ImageEtagExpected()) != 0`.
We didn't need this check before because we couldn't reach this point unless we set `If-None-Match`, but now we can.

DarthSim (changes requested):
Thanks for your contribution! There are a couple of things that need to be added before I can merge this:

1. Add the `IMGPROXY_USE_LAST_MODIFIED` config, disabled by default. `If-Modified-Since` can't take config and processing options changes into account, so users may not want to use it. Thus we shouldn't use it by default.
2. Add `If-Modified-Since` support to transports (S3, GCS, etc).

processing_handler.go
ehiggs: done!

processing_handler.go
ehiggs: done!

ehiggs:
> Add the IMGPROXY_USE_LAST_MODIFIED config, disabled by default. If-Modified-Since can't take config and processing options changes into account, so users may not want to use it. Thus we shouldn't use it by default.

Done.

processing_handler.go
```diff
@@ -91,6 +91,14 @@ func setCacheControl(rw http.ResponseWriter, force *time.Time, originHeaders map
 	}
 }
 
+func setLastModified(rw http.ResponseWriter, originHeaders map[string]string) {
+	if originHeaders != nil {
```
DarthSim, line 95: We should probably add a `config.LastModifiedEnabled` check here. If we set the `Last-Modified` header, a client may use it for its cache revalidation. And that may lead to unpredictable behavior. Better safe than sorry.

BTW, it's safe to get access to nil map's keys, so we don't need the `originHeaders != nil` check: https://goplay.tools/snippet/3OsmlN7HqxD

DarthSim:
@ehiggs I refactored ETag support in transports so it should be easier to add `Last-Modified` support: https://github.com/imgproxy/imgproxy/commit/49e5eb063c65721e62f9cae980b875e35dea6cb6

processing_handler.go
ehiggs: Fixed!

transport/notmodified/notmodified.go
```diff
@@ -15,6 +16,27 @@ func Response(req *http.Request, header http.Header) *http.Response {
...
+			return nil
+		}
+		lastModified, err := time.Parse(http.TimeFormat, lastModifiedRaw)
+		if err != nil {
+			return nil
+		}
+		ifModifiedSinceRaw := req.Header.Get("If-Modified-Since")
+		if len(ifModifiedSinceRaw) == 0 {
+			return nil
+		}
```
DarthSim, line 31: Let's move this up so we check the `If-Modified-Since` existence before parsing `Last-Modified`

DarthSim:
Could you add `If-Modified-Since` support to the rest of the transports?

ehiggs-deliverect:
> Could you add If-Modified-Since support to the rest of the transports?

Sure. Are we ok with the fact that we format the time to a string in the GCS Object. Then we parse it again in the notmodified module to compare it with the parsed `If-Not-Modified` time. It seems wasteful to go back and forth between formats, but I don't want to go on refactoring adventures here.

transport/notmodified/notmodified.go
ehiggs: nice catch. Fixed!

ehiggs:
>Could you add If-Modified-Since support to the rest of the transports?

Should be done with tests now. 

Note: gofakes3 doesn't support Last-Modified.

DarthSim:
LGTM! I'll try to patch gofakes3 before merging this. Doesn't look like a big deal.

ehiggs:
> LGTM! I'll try to patch gofakes3 before merging this. Doesn't look like a big deal.

Do you have a link to your upstream ticket? Thanks!

DarthSim:
I updated the master branch to use my fork of gofakes3. Could you uncomment the test?

ehiggs:
Done!

DarthSim:
Great job, Ewan! Thanks for your contribution!

PR #1152, opened 2023-05-04
allow acquiring s3 access credentials through role assumption

adomaskizogian:
With this change I introduce means to override IAM role by ussing sts assume role.

This allows to access S3 buckets by ECS or any other services that run on different accounts and helps to maintain good IAM security practices. Prior to this a credential pair was required to be made a available to the imgproxy service.

### DIFF
--- a/docs/serving_files_from_s3.md
+++ b/docs/serving_files_from_s3.md
@@ -6,7 +6,8 @@ imgproxy can process images from S3 buckets. To use this feature, do the followi
 2. [Set up the necessary credentials](#set-up-credentials) to grant access to your bucket.
 3. _(optional)_ Specify the AWS region with `IMGPROXY_S3_REGION` or `AWS_REGION`. Default: `us-west-1`
 4. _(optional)_ Specify the S3 endpoint with `IMGPROXY_S3_ENDPOINT`.
-5. Use `s3://%bucket_name/%file_key` as the source image URL.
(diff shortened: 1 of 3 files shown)


docs/serving_files_from_s3.md
```diff
@@ -49,6 +50,10 @@ aws_access_key_id = %access_key_id
 aws_secret_access_key = %secret_access_key
 ```
 
+#### Cross-Account Access
+
+S3 access credentials may be acquired by assuming a role using STS. To do so specify the IAM Role arn with `IMGPROXY_S3_ASSUME_ROLE_ARN` environment variable. This approach still requires you to provide initial AWS credentials by using one of the ways described above.
```
DarthSim, line 55:
```suggestion
S3 access credentials may be acquired by assuming a role using STS. To do so specify the IAM Role arn with the `IMGPROXY_S3_ASSUME_ROLE_ARN` environment variable. This approach still requires you to provide initial AWS credentials by using one of the ways described above. The provided credentials role should allow assuming the role with provided ARN.
```

DarthSim (changes requested):

docs/serving_files_from_s3.md
adomaskizogian: good point mentioning requirement for policy allowing to assume roles

DarthSim:
Thanks for your contribution!

PR #1155, opened 2023-05-11
Fix typo in presets docs

denisSurkov:
Hello!

Thank you for imgproxy and for your work. 

Just a small typo in presets documentation.

----

By the way, I think it's good to mention in presets docs, that presets are better approach if you want to save imgproxy url in database. 

The reason is if a developer wants to change all thumbnails (just for example) urls he might just change preset without processing urls in database.

I didn't put this description in this PR because I'm not sure is it really "good to mention" or everyone know it anyway.

### DIFF
--- a/docs/presets.md
+++ b/docs/presets.md
@@ -30,4 +30,4 @@ Setting `IMGPROXY_ONLY_PRESETS` to `true` switches imgproxy into "presets-only m
 http://imgproxy.example.com/unsafe/thumbnail:blurry:watermarked/plain/http://example.com/images/curiosity.jpg@png
 ```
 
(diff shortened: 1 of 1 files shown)


denisSurkov:
@DarthSim , sorry to bother you, can you review please?

It's strange I cannot tag anyone in reviewers field.

DarthSim:
Thanks for the fix!

I don't think that the way of using presets that you described is generally a good idea per se. URL generation doesn't take much time, so IMHO it's better to generate them on the fly instead of saving them in a database. Also, if you cache responses from imgproxy (which is generally recommended), you'll need to invalidate the cache after changing presets.

PR #1156, opened 2023-05-16
Drop dependency + use built-in `base64url`

nikeee:
[`create-hmac`](https://github.com/browserify/createHmac) is meant for compat between node and browser compat and was last updated in 2018.

Computing the HMAC signature on the client (browser) does not make any sense in 99% of the use-cases, as the secrets would be needed on the client side. This means that we can drop browser support and just use the native node module, which is also exported by the `create-hmac` when running on node.

`Buffer.toString()` also accepts "base64url" as an encoding, so we can drop the `urlSafeBase64` in favor of that.

### DIFF
--- a/examples/signature.js
+++ b/examples/signature.js
@@ -1,11 +1,7 @@
-const createHmac = require('create-hmac')
+import { createHmac } from 'node:crypto';
 
 const KEY = '943b421c9eb07c830af81030552c86009268de4e532ba2ee2eab8247c6da0881'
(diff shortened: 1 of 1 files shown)


DarthSim:
Hi @nikeee!

That looks nice! We made this example kinda universal so it could work in a browser too, but you're absolutely right about the meaninglessness of signing URLs on the client side.

Thanks for your contribution!

PR #1166, opened 2023-06-21
add query params support

tufitko:
What do you think about using query params?
This format is easiest to invalidate on CDNs. 
Signed requests will be able to invalidate too if a signature is in headers, not at the start of requests (I can make another pull request with this feature)

Examples of requests: 
```
http://imgproxy.example.com/insecure/aHR0cDovL2V4YW1w/bGUuY29tL2ltYWdl/cy9jdXJpb3NpdHku/anBn.png?rs=fill:300:400:0&g=sm

presets:
http://imgproxy.example.com/insecure/aHR0cDovL2V4YW1w/bGUuY29tL2ltYWdl/cy9jdXJpb3NpdHku/anBn.png?preset1&preset2
...
```

It's a simple draft and I'm new with imgproxy, so any questions and request's changes are welcome

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -36,7 +36,8 @@ var (
 
 	SoReuseport bool
 
-	PathPrefix string
+	PathPrefix     string
+	UseQueryParams bool
 
 	MaxSrcResolution            int
 	MaxSrcFileSize              int
@@ -229,6 +230,7 @@ func Reset() {
 	SoReuseport = false
 
 	PathPrefix = ""
+	UseQueryParams = false
 
 	MaxSrcResolution = 16800000
 	MaxSrcFileSize = 0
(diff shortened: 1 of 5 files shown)


tufitko:
@DarthSim what do you think about it? Is it possible to use query params?

DarthSim:
Hi @tufitko!

Thanks for the effort, but I can't accept this change. Mainly because it breaks one of the main imgproxy's principles – simplicity:

1. Query params can't be signed by design. First of all, we want to keep this territory free for use by others. But also, signing query params is pretty hard. Take a look at AWS Signature Version 4 and you'll know what I mean. And I really don't want to make it harder than it is.

2. Query params support means another URL format. We had two URL formats in v2 and we didn't like it.

Also, as you pointed out yourself, this change doesn't really make it easier to invalidate the URL's cache because of the signature. The signature should be in the URL because pitting it into headers won't allow using imgproxy URLs in the  tag. And again, this means additions to the URL format, lots of confused users, etc. Not the things we're looking for.

PR #1168, opened 2023-06-22
fix(svg/sanitize): preserve headers from origin data

tufitko:
`IMGPROXY_CACHE_CONTROL_PASSTHROUGH` doesn't work with `IMGPROXY_SANITIZE_SVG`

after calling `Sanitize`, cache-control headers are disappeared, so we need to preserve origin headers while creating a new `ImageData`


tufitko:
@DarthSim can you check this?

DarthSim:
Dang, I was going to merge this to v3.18.1 but completely forgot 😓
I merged this manually with some additions. Thanks for your contribution!

PR #1169, opened 2023-06-25
Fix Java example and add Clojure example

apeckham:


DarthSim:
Thanks for your contribution!

PR #1185, opened 2023-08-06
Fix typo in cloud_watch.md

PhantomRay:


DarthSim:
Ouch! Thanks for the fix!

PR #1200, opened 2023-09-07
Fix to add null pointer in vips_copy method arguments

mbanusic:
When compiling from source on Ubuntu we get an error: 

# github.com/imgproxy/imgproxy/v3/vips
vips.c: In function ‘vips_fix_float_tiff’:
vips.c:247:3: warning: not enough variable arguments to fit a sentinel [-Wformat=]
  247 |   return vips_copy(in, out);
      |   ^~~~~~

vips_copy has more arguments and call needs to be terminated with null pointer.

This fixes that problem.


DarthSim:
Dang, how didn't I catch this? Thanks a lot!

Issue #1208, opened 2023-10-03
Support for S3 client side encryption?


gjtorikian:
I took a quick look through the S3 portion of the project and didn’t see support for [client side encryption](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingClientSideEncryption.html). Is this something you’d consider adding?

Looking through my own S3 client code, I believe all an end user would need to provide is their KMS key:

```
kms_key_id: kms_key_id,
key_wrap_schema: :kms_context,
content_encryption_schema: :aes_gcm_no_padding,
security_profile: :v2,
```

The rest of the values are just defaults.

I’d be happy to start a PR to add this if there isn’t already work for it underway.

gjtorikian:
closed by https://github.com/imgproxy/imgproxy/pull/1210

PR #1210, opened 2023-10-12
Support encrypted client objects in S3 buckets

gjtorikian:
👋 @DarthSim I didn't hear back about https://github.com/imgproxy/imgproxy/issues/1208, and I needed this feature, so I've added it. Obviously, this PR is missing tests and documentation, but I wanted to first run this by you before putting more work into it. I am also not sure if this repo is watched/maintained as some PRs (like https://github.com/imgproxy/imgproxy/pull/1196) have been open for a while.

Regardless, the changes here fit my use case! I think the only open question I have is _how_ to toggle between a regular client and a decrypted one. There are two choices:

1. Issue one request with the default client; if certain headers suggest that the object is encrypted, issue a second request with the decryption client; or
2. Always use the decryption client. This can be toggled with an env var like `IMGPROXY_S3_USE_CLIENT_ENCRYPTION` (or whatever name you like).

Option 1 offers more flexibility at the cost of another HTTP request; option 2 is best suited for situations like mine, where I know every object in the bucket is encrypted. What do you think?

Will close https://github.com/imgproxy/imgproxy/issues/1208 if merged.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -99,11 +99,12 @@ var (
 
 	LocalFileSystemRoot string
 
-	S3Enabled       bool
-	S3Region        string
-	S3Endpoint      string
-	S3AssumeRoleArn string
-	S3MultiRegion   bool
+	S3Enabled                 bool
+	S3Region                  string
+	S3Endpoint                string
+	S3AssumeRoleArn           string
+	S3MultiRegion             bool
+	S3DecryptionClientEnabled bool
 
 	GCSEnabled  bool
 	GCSKey      string
@@ -300,6 +301,7 @@ func Reset() {
 	S3Endpoint = ""
 	S3AssumeRoleArn = ""
 	S3MultiRegion = false
+	S3DecryptionClientEnabled = false
 	GCSEnabled = false
 	GCSKey = ""
 	ABSEnabled = false
@@ -501,6 +503,7 @@ func Configure() error {
 	configurators.String(&S3Endpoint, "IMGPROXY_S3_ENDPOINT")
 	configurators.String(&S3AssumeRoleArn, "IMGPROXY_S3_ASSUME_ROLE_ARN")
 	configurators.Bool(&S3MultiRegion, "IMGPROXY_S3_MULTI_REGION")
+	configurators.Bool(&S3DecryptionClientEnabled, "IMGPROXY_S3_USE_DECRYPTION_CLIENT")
 
 	configurators.Bool(&GCSEnabled, "IMGPROXY_USE_GCS")
 	configurators.String(&GCSKey, "IMGPROXY_GCS_KEY")
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -371,6 +371,7 @@ imgproxy can process files from Amazon S3 buckets, but this feature is disabled 
 * `IMGPROXY_S3_REGION`: an S3 buckets region
 * `IMGPROXY_S3_ENDPOINT`: a custom S3 endpoint to being used by imgproxy
 * `IMGPROXY_S3_MULTI_REGION`: when `true`, allows using S3 buckets from different regions
+* `IMGPROXY_S3_USE_DECRYPTION_CLIENT`: when `true`, toggles client-side encryption
 
 Check out the [Serving files from S3](serving_files_from_s3.md) guide to learn more.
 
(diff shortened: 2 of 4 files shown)


DarthSim:
Hey @gjtorikian!

Sorry for the delay, there's a lot on my plate right now.

First of all, thanks for the effort. However, this PR has some problems that need to be fixed before I can merge this.

1. The decryption should be disabled by default. I'm not sure that the decryption client won't fail when an object is not encrypted. Anyway, this involves additional actions most users don't need. The client creation may be wrapped into a function that returns `s3.S3` or `s3crypto.DecryptionClientV2` depending on if decryption is enabled or disabled. Both types may be represented as an interface implementing `GetObjectRequest`.
2. `s3crypto.NewDecryptionClientV2` creates it's own S3 client without a way to provide its config. This breaks `IMGPROXY_S3_ASSUME_ROLE_ARN` and `IMGPROXY_S3_MULTI_REGION`. Since `s3crypto.NewDecryptionClientV2` accepts only `ConfigProvider`, we could wrap `Session` into a custom implementation of `ConfigProvider` that adds a provided `Config` to `ConfigProvider.ClientConfig` args. This is a pretty dirty way, but `NewDecryptionClientV2` leaves us no choice.
3. The decryption client should be returned by `getClient` if decryption is enabled.
4. The crypto registry should be initialized once on the transport creation.
5. `X-Amz-Meta-X-Amz-Unencrypted-Content-Length` should be handled by the transport.

gjtorikian:
Right, all your feedback is spot on. I intended for this to be a spike to show how minimal the change is.

> depending on if decryption is enabled or disabled

The crux of my question in the PR is the same as what you mentioned: how _should_ this client be toggled? I guess I’ll rely on another environment variable? 

Once I have a path there, of course I will make sure to make the decryption client toggleable and abstracted in the ways you suggest.  👍

DarthSim:
> I guess I’ll rely on another environment variable?

Yup

gjtorikian:
Ah, I guess now is time to admit that Go is not a language I code in regularly. 😓 

I incorporated your feedback as best as I could. I think one thing to note is that

> `s3crypto.NewDecryptionClientV2` creates it's own S3 client without a way to provide its config.

I think that the previous spike didn't show that [you can pass in a "base client" to the decryption client](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3crypto/#DecryptionClientOptions).  In this case we can always create the `defaultClient` and provide it to the decryption client if it's toggled. When it's time to fetch the object, we do so based on one or the other client. So far as I know, this still would support multi-region.

Prior to what I've committed, I did try to work with an interface like this:

```go
type S3Client interface {
	GetObjectRequest(input *s3.GetObjectInput) (req *request.Request, output *s3.GetObjectOutput)
}

type S3DefaultClient struct {
	client *s3.S3
}

type S3DecryptionClientV2 struct {
	decryptionClient *s3crypto.DecryptionClientV2
}


func (defaultClient S3DefaultClient) GetObjectRequest(input *s3.GetObjectInput) (req *request.Request, output *s3.GetObjectOutput) {
	return defaultClient.GetObjectRequest(input)
}

func (decryptionClient S3DecryptionClientV2) GetObjectRequest(input *s3.GetObjectInput) (req *request.Request, output *s3.GetObjectOutput) {
  // etc
}
```

But my Go isn't good enough to figure out what had been going wrong in this direction. If you prefer it, I can go in that direction instead.

I'll add tests once the final architecture is approved!

DarthSim:
Both `*s3.S3` and `decryptionClient *s3crypto.DecryptionClientV2` already implement this interface:

```go
type S3Client interface {
	GetObjectRequest(input *s3.GetObjectInput) (req *request.Request, output *s3.GetObjectOutput)
}
```

...so there's no need to wrap them.

> I think that the previous spike didn't show that [you can pass in a "base client" to the decryption client](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3crypto/#DecryptionClientOptions)

Well, yes, but AWS SDK configures the client for decryption under the hood of `s3crypto.NewDecryptionClientV2`, so I would rather not provide a custom client to it. Instead, we could do the following:

1. Add `defaultConfig` to `transport` and save `conf` to it during the transport creation.
2. In `getClient` (or better extract this into `createClient`):
  ```go
  conf := t.defaultClient.Config.Copy()
  conf.Region = aws.String(region)
  
  sess := t.sess.Copy(conf)
  
  // s3Client is the interface described above
  var (
    client s3Client
    err error
  )
  
  if config.S3ClientEncryptionEnabled {
    client, err = s3crypto.NewDecryptionClientV2(session, cr)
  } else {
    client, err = s3.New(t.session)
  }
  ```

DarthSim:
Hi @gjtorikian!
I polished this a bit and merged it to master manually: https://github.com/imgproxy/imgproxy/commit/b384e2bb7f997655f720ce1637284f4725de964d
Thanks for your contribution! 👍

gjtorikian:
Ah perfect, thank you so much! I was having trouble writing the test code.

PR #1215, opened 2023-10-25
Added example of usage in CloudFront Functions

levnovikov:
Modified signature.js for use in AWS CloudFront Functions.

About JS runtime and restrictions - 
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/functions-javascript-runtime-features.html


DarthSim:
Thanks for your contribution! 👍

Issue #1211, opened 2023-10-16
Documentation: Server requirements/recomendations 


schonert:
We're moving around a bit and would like to see if there would be any gains in switching to a different server type.
It would be great to have a bit of documentation on requirements and recommendations.

With recommendations, I'm curious about:
1) Size vs count. Would 4x 2cpu/2gp servers be better than a 8cpu/8gp. 
2) CPU vs RAM vs GPU. What should we go for and would there be any gains with a GPU?

DarthSim:
Hey @schonert!

TL;DR: it's better to have multiple small instances than a single big one. The rule of thumb is 2 CPUs per instance and 2 workers (concurrency) per CPU. If the average downloading time is much less than the average processing time, then it makes sense to lower workers per CPU number and spin out more instances.

imgproxy itself and its dependencies have multiple global mutexes under the hood. Also, the system threads scheduler is not completely free. Thus, multiple processes perform better than a single process. The best setup for a uniform load is 1 instance per CPU. However, most load balancers are pretty stupid and use the round-robin algorithm. This may lead to a situation where an instance is busy processing a heavy request but the load balancer routes another request to it, so it's better to have a spare CPU.

imgproxy doesn't use GPU, so there's no sense in spending money on it.

The memory size doesn't affect performance either. Just use enough memory to store the source image file, decoded source image, and the resulting image file, and it should be fine. If you use ML options, you'll also need enough memory to fit neural network models. The models distributed with imgproxy Pro are pretty small though. Since you have been running imgproxy for a pretty long time, you are probably pretty aware of how much memory it uses in your case.

What really affects performance is the type of CPU:
1. Shared CPU vs dedicated CPU. Many hostings offer machines with both shared and dedicated CPUs. The performance of shared CPUs is unpredictable and usually significantly lower than that of dedicated CPUs. In the case of AWS, I'd very much recommend utilizing c-type EC2 instances rather than t-type.
2. AMD64 vs ARM64. I'm not very familiar with other platforms, but AWS offers absolutely amazing c7g instances that are using ARM64 Amazon Graviton 3 CPUs. In our tests, Amazon Graviton 3 outperforms CPUs used in c7a instances a lot.

schonert:
Brilliant reply @DarthSim - appreciated!
I'm sure this will be useful for others - so we could consider adding it to the doc site.

[...]

PR #1217, opened 2023-10-27
Add one-click deploy to docs

jbgh:
@DarthSim I saw your comment here: https://github.com/imgproxy/imgproxy/issues/1211#issuecomment-1766680977 and thought that we could help provide a cool one-click deploy experience for imgproxy users.

We’ve made the default [one-click deployment](https://app.trydome.io/signup?package=imgproxy) 0.75 CPU cores + 1GB of RAM and it seems to be working fine but could up it to the 2 CPU cores that you recommended for the reasons you stated (multiple global mutexes + round robin LBs potentially sending another job to an already busy instance).

Dome is pretty flexible so it’s possible for users to configure high CPU core counts and still stick with 1GB of RAM or you could also go horizontal with as many 2 CPU cores + 256MB RAM replicas as needed if for example the use case is high concurrency processing of relatively small sized images.  Let me know if you have any additional thoughts.  Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ In terms of security, the massive processing of remote images is a potentially d
 
 Check out our 📑 [Documentation](https://docs.imgproxy.net).
 
+## Deploy
+
+Deploy imgproxy to Dome in [one-click](https://app.trydome.io/signup?package=imgproxy):
+
+[![Deploy to Dome](https://trydome.io/button.svg)](https://app.trydome.io/signup?package=imgproxy)
+
 ## Author
 
 Sergey "[DarthSim](https://github.com/DarthSim)" Alexandrovich
--- a/README.md
+++ b/README.md
@@ -79,12 +79,6 @@ In terms of security, the massive processing of remote images is a potentially d
 
 Check out our 📑 [Documentation](https://docs.imgproxy.net).
 
-## Deploy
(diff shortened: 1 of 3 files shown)


DarthSim:
Hi @jbgh!

We have a documentation page about installation ways: https://github.com/imgproxy/imgproxy/blob/master/docs/installation.md. It's better to add the button there rather than to  README

jbgh:
Hi @DarthSim thanks for the feedback!  I've moved the button to your docs and removed it from the readme.

jbgh:
@DarthSim bumping for review.  Let me know if you have any other thoughts/feedback.  Thanks!

DarthSim:
We moved documentation source code to https://github.com/imgproxy/imgproxy-docs. I already added a button for Dome: https://docs.imgproxy.net/installation#dome

jbgh:
Thanks @DarthSim for the merge!  Reach out to jay@trydome.io if you have any requests or issues.

jbgh:
@DarthSim looks like the link in the docs is incorrect.  The link on `Dome` in the docs is pointing to heroku.  This is the correct link:

https://app.trydome.io/signup?package=imgproxy

DarthSim:
My bad. Fixed

Issue #1251, opened 2024-03-05
SVG source format doesn't work with WEBP_DETECTION and AVIF_DETECTION


Frezyy123:
Hi, I am facing with some undocumented behaviour
If I have svg image as a source and enable all features with [AVIF/WebP support detection](https://docs.imgproxy.net/configuration/options#avifwebp-support-detection), svg image will not be processed and imgproxy return svg image.

Looks like it skips everytime, when we have svg source image with missing extension https://github.com/imgproxy/imgproxy/blob/5d4a612db79f824f3acd93954f7f30c17aef6a4e/processing_handler.go#L371-L375

So, everytime, when I have svg source, I need to add any extension, just to force this feature with detection. For example
https://example.com/image `returns svg image`
https://example.com/image@jpeg `returns in respect of accept header, webp or avif`

Could you please explain to me, why it works like this? Is there any problem with conversion svg?

DarthSim:
Hey @Frezyy123!

SVGs are not rasterized unless the resulting format is explicitly specified. In the vast majority of cases rasterizing SVG doesn't bring any benefit. The only case I know where it makes sense is email.

However, having a `rasterize_svg` option may be useful 🤔

Frezyy123:
In some cases svg icon is much more easier to draw, but iOS application have troubles with svg and choose webp format (or pdf but this is really bad way :))

I will try to implement this option and bring pull request to you
Thank you

PR #1257, opened 2024-03-15
feat: Implement AlwaysRasterizeSvg

Frezyy123:
https://github.com/imgproxy/imgproxy/issues/1251
* Add EnableForceRasterizeSVG option
* If enabled it skips conversion svg--->svg with unknown requested format

### DIFF
--- a/processing_handler.go
+++ b/processing_handler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -369,32 +370,28 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 
 	checkErr(ctx, "timeout", router.CheckTimeout(ctx))
 
-	if originData.Type == po.Format || po.Format == imagetype.Unknown {
-		// Don't process SVG
-		if originData.Type == imagetype.SVG {
-			if config.SanitizeSvg {
-				sanitized, svgErr := svg.Sanitize(originData)
-				checkErr(ctx, "svg_processing", svgErr)
+	// Skip processing svg with unknown or the same destination imageType
+	// if it's not forced by AlwaysRasterizeSvg option
+	// Also skip processing if the format is in SkipProcessingFormats
+	shouldSkipProcessing := (originData.Type == po.Format || po.Format == imagetype.Unknown) &&
+		(slices.Contains(po.SkipProcessingFormats, originData.Type) ||
+			originData.Type == imagetype.SVG && !config.AlwaysRasterizeSvg)
 
-				// Since we'll replace origin data, it's better to close it to return
-				// it's buffer to the pool
-				originData.Close()
+	if shouldSkipProcessing {
+		if originData.Type == imagetype.SVG && config.SanitizeSvg {
+			sanitized, svgErr := svg.Sanitize(originData)
+			checkErr(ctx, "svg_processing", svgErr)
 
-				originData = sanitized
-			}
+			// Since we'll replace origin data, it's better to close it to return
+			// it's buffer to the pool
+			originData.Close()
 
-			respondWithImage(reqID, r, rw, statusCode, originData, po, imageURL, originData)
-			return
+			originData = sanitized
+
 		}
 
-		if len(po.SkipProcessingFormats) > 0 {
-			for _, f := range po.SkipProcessingFormats {
-				if f == originData.Type {
-					respondWithImage(reqID, r, rw, statusCode, originData, po, imageURL, originData)
-					return
-				}
-			}
-		}
+		respondWithImage(reqID, r, rw, statusCode, originData, po, imageURL, originData)
+		return
 	}
 
 	if !vips.SupportsLoad(originData.Type) {
--- a/processing_handler_test.go
+++ b/processing_handler_test.go
@@ -608,7 +608,6 @@ func (s *ProcessingHandlerTestSuite) TestModifiedSinceReqExactMatchLastModifiedD
 		require.Equal(s.T(), "", modifiedSince)
 		rw.WriteHeader(200)
 		rw.Write(data)
-
 	}))
 	defer ts.Close()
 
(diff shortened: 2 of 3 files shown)


DarthSim:
Hey @Frezyy123!

Thanks for the PR!

There's a case that is not covered: if `EnableForceRasterizeSVG` is `true` but `po.SkipProcessingFormats` includes `svg`, an SVG image won't be rasterized (as expected) yet won't be sanitized and minimized.

BTW, `po.SkipProcessingFormats` check can be replaced with a one-liner: `slices.Contains(po.SkipProcessingFormats, originData.Type)`

Also, I'd name the config `AlwaysRasterizeSvg` and put it after `SanitizeSvg`.

schonert:
Wouldn't this be solved by simply setting the format option?

DarthSim:
If you know the original format, yeah, it will. Otherwise, you probably don't want to convert your JPEGs to PNGs

Frezyy123:
I renamed option to AlwaysRasterizeSvg, replaced skipProcessingOptions and added sanitize everywhere
Also i wrote simple tests for some cases that cover this option

Frezyy123:
btw, I see in documentation that SkipProcessing works only when the requested format is the same. But when we have unknown type it works too. 
```
Processing can only be skipped when the requested format is the same as the source format.

```
https://docs.imgproxy.net/usage/processing#skip-processing

What do you think should we divide these conditions? Or maybe we need to fix documentation?
```
if originData.Type == po.Format || po.Format == imagetype.Unknown {
```

processing_handler.go
```diff
@@ -370,8 +371,9 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
 	checkErr(ctx, "timeout", router.CheckTimeout(ctx))
 
 	if originData.Type == po.Format || po.Format == imagetype.Unknown {
-		// Don't process SVG
-		if originData.Type == imagetype.SVG {
+		// Skip processing svg with unknown destination imageType
+		// if it's not forced by AlwaysRasterizeSvg option
+		if originData.Type == imagetype.SVG && !config.AlwaysRasterizeSvg {
```
DarthSim, line 376: It's better to check `po.SkipProcessingFormats` here instead of duplicating the whole block further

```suggestion
		shouldSkip := slices.Contains(po.SkipProcessingFormats, originData.Type)

		// Skip processing svg with unknown destination imageType
		// if it's not forced by AlwaysRasterizeSvg option
		if originData.Type == imagetype.SVG && (!config.AlwaysRasterizeSvg || shouldSkip) {
```

processing_handler_test.go
```diff
@@ -721,6 +724,73 @@ func (s *ProcessingHandlerTestSuite) TestModifiedSinceReqCompareTooOldLastModifi
...
+	rw := s.send("/unsafe/rs:fill:40:40/plain/local:///test1.svg")
+	res := rw.Result()
+
+	require.Equal(s.T(), 200, res.StatusCode)
+	require.Equal(s.T(), "image/png", res.Header.Get("Content-Type"))
+
+	actual := s.readBody(res)
+	expected := s.readTestFile("test1.fromsvg.png")
+
+	require.True(s.T(), bytes.Equal(expected, actual))
```
DarthSim, line 740: It's better to just check `Content-Type` header. Otherwise, this test may fail because of changes in librsvg or libspng. Same for the rest of the tests

DarthSim (changes requested):

processing_handler.go
Frezyy123: It's not quite right condition and it fails even with old tests, but i try to rewrite it. 
However there are still duplicating with svg case
```golang
if originData.Type == imagetype.SVG && !config.AlwaysRasterizeSvg || shouldSkipProcessing {
			if originData.Type == imagetype.SVG {
```
We have two cases, svg origin data with disabled AlwaysRasterizeSvg and skipProcessing with any origin data. So our condition should work only in both cases
And then we should sanitize svg, so we have to put another check on svg and skip other formats

processing_handler_test.go
Frezyy123: You are right, I droped the requires and test data that I added

processing_handler.go
DarthSim: This is a valid condition, you just didn't use it the way I meant. But that's ok

processing_handler_test.go
```diff
@@ -721,6 +724,49 @@ func (s *ProcessingHandlerTestSuite) TestModifiedSinceReqCompareTooOldLastModifi
...
+	rw := s.send("/unsafe/rs:fill:40:40/plain/local:///test1.svg")
+	res := rw.Result()
+
+	require.Equal(s.T(), 200, res.StatusCode)
+	require.Equal(s.T(), "image/png", res.Header.Get("Content-Type"))
+}
+
+func (s *ProcessingHandlerTestSuite) TestAlwaysRasterizeSvgWithEnforceAvif() {
+	config.AlwaysRasterizeSvg = true
+	config.EnforceAvif = true
```
DarthSim, line 740: Let's change this to WebP. The CI Docker image doesn't include libheif

DarthSim: And in other tests too

processing_handler.go
```diff
@@ -370,31 +371,28 @@ func handleProcessing(reqID string, rw http.ResponseWriter, r *http.Request) {
...
+					// it's buffer to the pool
+					originData.Close()
+
+					originData = sanitized
+				}
 			}
 
 			respondWithImage(reqID, r, rw, statusCode, originData, po, imageURL, originData)
 			return
 		}
```
DarthSim, line 395: If we're simplifying things here, we could've simplify 'em even more 🙂

```go
shouldSkipProcessing := (originData.Type == po.Format || po.Format == imagetype.Unknown) &&
	(slices.Contains(po.SkipProcessingFormats, originData.Type) ||
		originData.Type == imagetype.SVG && config.AlwaysRasterizeSvg)

if shouldSkipProcessing {
	if config.SanitizeSvg {
		// ... do SVG stuff
	}

	respondWithImage(reqID, r, rw, statusCode, originData, po, imageURL, originData)
	return
}
```

Frezyy123: done
Only thing that I changed is 
```
config.AlwaysRasterizeSvg --> !config.AlwaysRasterizeSvg
```

processing_handler_test.go
Frezyy123: Changed to webp

DarthSim:
Merged. Thanks for your contribution!

PR #1258, opened 2024-03-19
Add latency in nanoseconds as a response log field

tab1293:
This change adds the latency of a request as a response log field. This is useful when filtering logs in Google Cloud logging for requests that have high latencies.

### DIFF
--- a/router/logging.go
+++ b/router/logging.go
@@ -3,6 +3,7 @@ package router
 import (
 	"net"
 	"net/http"
+	"strconv"
 
 	"github.com/imgproxy/imgproxy/v3/ierrors"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


router/logging.go
```diff
@@ -33,12 +34,15 @@ func LogResponse(reqID string, r *http.Request, status int, err *ierrors.Error,
 	}
 
 	clientIP, _, _ := net.SplitHostPort(r.RemoteAddr)
+	ctxDuration := ctxTime(r.Context())
+	latency := strconv.FormatInt(int64(ctxDuration), 10)
```
DarthSim, line 38: There's no need to stringify `ctxDuration`, logrus works well with numbers.

Also, this will return the duration in nanoseconds which I see a little bit redundant. I believe millisecond precision should be enough.

DarthSim (changes requested):
Hey @tab1293!

Thanks for the PR! I'm ok with adding latency to the logs but take a look at my comments, please.

PR #1272, opened 2024-04-12
chore: fix some typos in comments

camcui:


DarthSim:
Hey @camcui!

The text in the `NOTICE` file should not be fixed. It contains license texts of imgproxy's dependencies, and they should be provided as-is.

DarthSim:
Thanks

PR #1290, opened 2024-05-16
Add support for passing in an external ID when using S3 assume role

### LINKED ISSUE #1289: Support for external IDs when using S3 with assumed roles
Currently, `imgproxy` does not support passing in an external ID when assuming a role for S3 backends. The external ID provides an extra layer of security. The change itself is pretty straight forward. If interested, I can open a PR to add this functionality in.

riyadhalnur:
- Adds support for passing in an optional external ID config parameter for S3 assume roles
- Will open a separate PR to update the documentation once this is merged in

Closes #1289

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -107,6 +107,7 @@ var (
 	S3Region                  string
 	S3Endpoint                string
 	S3AssumeRoleArn           string
+	S3AssumeRoleExternalID    string
 	S3MultiRegion             bool
 	S3DecryptionClientEnabled bool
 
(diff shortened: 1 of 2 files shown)


DarthSim:
Hey @riyadhalnur!

Thanks for the contribution! Everything looks good, but I doubt we need that warning in `config.go`. We'll cover this in the docs.

riyadhalnur:
Fair enough. Removed the warning log. I'll create a PR in the docs repo soon for this change.

DarthSim:
Merged! Thanks for your contribution 👍

PR #1291, opened 2024-05-19
Fix spelling of "project" in airbrake config keys

koenpunt:
Noticed the "t" was missing.


DarthSim:
Thanks for the catch!

PR #1293, opened 2024-05-24
Add support for padding values including zero in imgproxy

biodigitalfish:
### Description

This PR adds support for specifying padding values including zero in imgproxy. Previously, padding values of zero were not accepted, which limited the flexibility of the padding option. Additionally, padding will only be enabled if at least one padding value is greater than zero.

### Changes

1. Modified the `applyPaddingOption` function to correctly handle padding values including zero.
2. Enabled the `po.Padding` flag based on the presence of at least one padding value greater than zero.

### Modified Code

```go
func applyPaddingOption(po *ProcessingOptions, args []string) error {
	nArgs := len(args)

	if nArgs < 1 || nArgs > 4 {
		return fmt.Errorf("Invalid padding arguments: %v", args)
	}

	if nArgs > 0 {
		if err := parseDimension(&po.Padding.Top, "padding top (+all)", args[0]); err != nil {
			return err
		}
		po.Padding.Right = po.Padding.Top
		po.Padding.Bottom = po.Padding.Top
		po.Padding.Left = po.Padding.Top
	}

	if nArgs > 1 {
		if err := parseDimension(&po.Padding.Right, "padding right (+left)", args[1]); err != nil {
			return err
		}
		po.Padding.Left = po.Padding.Right
	}

	if nArgs > 2 {
		if err := parseDimension(&po.Padding.Bottom, "padding bottom", args[2]); err != nil {
			return err
		}
	}

	if nArgs > 3 {
		if err := parseDimension(&po.Padding.Left, "padding left", args[3]); err != nil {
			return err
		}
	}

	// Enable padding only if at least one padding value is greater than 0
	po.Padding.Enabled = po.Padding.Top > 0 || po.Padding.Right > 0 || po.Padding.Bottom > 0 || po.Padding.Left > 0

	return nil
}

### DIFF
--- a/options/processing_options.go
+++ b/options/processing_options.go
@@ -480,9 +480,7 @@ func applyPaddingOption(po *ProcessingOptions, args []string) error {
 		return fmt.Errorf("Invalid padding arguments: %v", args)
 	}
 
-	po.Padding.Enabled = true
-
-	if nArgs > 0 && len(args[0]) > 0 {
+	if nArgs > 0 {
 		if err := parseDimension(&po.Padding.Top, "padding top (+all)", args[0]); err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


biodigitalfish:
@DarthSim - it doesn't work for me when i try do something like this:
`padding: {top:10, bottom:10, right: 0, left:0 }`

But it works if i do:
`padding: {top:10, bottom:10, right:1, left:1}`

So unless i'm doing something wrong it doesn't accept 0 values in this way?

DarthSim:
You can do this with `pd:10:0` or `pd:10:0:10:0`

biodigitalfish:
aaah. i think the issue might actually be with imgproxy-js-core evaluating 0 value arg as falsey. I'm using the imgproxy-node package. Should have mentioned. 
Should i raise a PR for this block? 
https://github.com/imgproxy/imgproxy-js-core/blob/24731f12a1a8b81afae62d6b261035506f2d156e/src/options/padding.ts#L22C1-L30C39

PR #1410, opened 2024-12-10
Fix ld: error: undefined symbol: ceil on FreeBSD

tmaroschik:
When building on FreeBSD the compilation errors because it cannot find the math library. This single line links it.


DarthSim:
Hey @tmaroschik!
Thanks for the contribution!

PR #1411, opened 2024-12-29
Fix question mark handling for non-http URLs

naglfar:
Recent releases saw some changes to how special characters are handled in non-HTTP URLs, particularly a special case was added for ```#```. Since local URLs / files may include ```?``` as well (at least on Linux) this needs to be handled as well.


DarthSim:
I added the [IMGPROXY_SOURCE_URL_QUERY_SEPARATOR](https://docs.imgproxy.net/latest/configuration/options#IMGPROXY_SOURCE_URL_QUERY_SEPARATOR) config to the `latest` build. It allows to mitigate the issue.

PR #1413, opened 2025-02-10
add query string passthrough and disable cookie rfc checks

tonybart1337:
We found it cumbersome to encode all URLs when they're passed through the imgproxy service. Since this encoding logic would have had to be implemented in every service that generates links for imgproxy, we introduced a new setting that allows us to pass the query string through as is.

Another difficulty is cookie passthrough, by default it checks headers according to the RFC, in our case clients can use different hosts to access the server, and we want to keep it easy for them configuration-wise, we control all the services that receive the cookies anyway, so not a concern for us

I'm not a go-master, but I try

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -102,10 +102,12 @@ var (
 	SanitizeSvg        bool
 	AlwaysRasterizeSvg bool
 
-	CookiePassthrough bool
-	CookieBaseURL     string
+	CookiePassthrough   bool
+	CookieDisableChecks bool
+	CookieBaseURL       string
 
-	SourceURLQuerySeparator string
+	SourceURLQuerySeparator   string
+	SourceURLQueryPassthrough bool
 
 	LocalFileSystemRoot string
 
@@ -561,7 +563,10 @@ func Configure() error {
 	configurators.Bool(&DevelopmentErrorsMode, "IMGPROXY_DEVELOPMENT_ERRORS_MODE")
 
 	configurators.Bool(&CookiePassthrough, "IMGPROXY_COOKIE_PASSTHROUGH")
+	configurators.Bool(&CookieDisableChecks, "IMGPROXY_COOKIE_DISABLE_RFC_CHECKS")
 	configurators.String(&CookieBaseURL, "IMGPROXY_COOKIE_BASE_URL")
+
+	configurators.Bool(&SourceURLQueryPassthrough, "IMGPROXY_SOURCE_URL_QUERY_PASSTHROUGH")
 
 	// Can't rely on configurators.String here because it ignores empty values
 	if s, ok := os.LookupEnv("IMGPROXY_SOURCE_URL_QUERY_SEPARATOR"); ok {
--- a/cookies/cookies.go
+++ b/cookies/cookies.go
@@ -5,16 +5,63 @@ import (
 	"net/http"
 	"net/http/cookiejar"
 	"net/url"
+	"sync"
 
 	"golang.org/x/net/publicsuffix"
 
 	"github.com/imgproxy/imgproxy/v3/config"
 )
 
-func JarFromRequest(r *http.Request) (*cookiejar.Jar, error) {
-	jar, err := cookiejar.New(&cookiejar.Options{PublicSuffixList: publicsuffix.List})
+// PassthroughCookieJar is a custom CookieJar that can return either all cookies
+// or only those according to the RFC6265 rules.
+type PassthroughCookieJar struct {
+	jar             http.CookieJar // underlying default jar (RFC6265 compliant)
+	alwaysReturnAll bool           // when true, Cookies() returns all stored cookies
+	mu              sync.Mutex     // protects allCookies and alwaysReturnAll
+	allCookies      []*http.Cookie // holds every cookie added via SetCookies
+}
+
+// Default implementation doesn't expose cookies, so we need to keep track of them in our slice
+func (ocj *PassthroughCookieJar) SetCookies(u *url.URL, cookies []*http.Cookie) {
+	ocj.jar.SetCookies(u, cookies)
+	ocj.mu.Lock()
+	// Simply append the new cookies.
+	ocj.allCookies = append(ocj.allCookies, cookies...)
+	ocj.mu.Unlock()
+}
+
+// Cookies returns cookies according to the runtime flag.
+// If alwaysReturnAll is true, it returns all stored cookies.
+// Otherwise, it returns the cookies for the given URL as computed by the underlying jar.
+func (ocj *PassthroughCookieJar) Cookies(u *url.URL) []*http.Cookie {
+	ocj.mu.Lock()
+	defer ocj.mu.Unlock()
+	if ocj.alwaysReturnAll {
+		// Return a copy of all cookies.
+		copied := make([]*http.Cookie, len(ocj.allCookies))
+		copy(copied, ocj.allCookies)
+		return copied
+	}
+	return ocj.jar.Cookies(u)
+}
+
+// SetAlwaysReturnAll lets you toggle the behavior at runtime.
+func (ocj *PassthroughCookieJar) SetAlwaysReturnAll(always bool) {
+	ocj.mu.Lock()
+	ocj.alwaysReturnAll = always
+	ocj.mu.Unlock()
+}
+
+func JarFromRequest(r *http.Request) (*PassthroughCookieJar, error) {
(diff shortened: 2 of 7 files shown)


DarthSim:
Hey @tonybart1337!

Thanks for the PR!

> We found it cumbersome to encode all URLs when they're passed through the imgproxy service. Since this encoding logic would have had to be implemented in every service that generates links for imgproxy, we introduced a new setting that allows us to pass the query string through as is.

Though I'm not against this feature, I'd want to point out that `?` is not the only symbol that requires percent encoding. `@` and `%` should also be encoded, so this feature doesn't solve the encoding problem completely.

The vast majority of languages have packages/functions for URL encoding, so I honestly don't see how this is a problem.

> Another difficulty is cookie passthrough, by default it checks headers according to the RFC, in our case clients can use different hosts to access the server, and we want to keep it easy for them configuration-wise, we control all the services that receive the cookies anyway, so not a concern for us

Could you give me an example that doesn't pass RFC checks?

DarthSim:
> We don't want to overcomplicate the code in all the services that might need to receive previews

If you're going to generate imgproxy URLs on the client side, you'll need some code for the URL generation anyway, right? So instead of (in JS)

```javascript
"/unsafe/rs:fit:100:100/" + imageURL
```

...you'll have something like

```javascript
"/unsafe/rs:fit:100:100/" + encodeURIComponent(imageURL)
```
Or, if you just want to encode `?`:

```javascript
"/unsafe/rs:fit:100:100/" + imageURL.replaceAll("?", "%3F")
```

That doesn't look like an overcomplication at all to me.

> are you sure it's required?

If your URLs may contain `%` or `@`, then yes, it is required.

Again, I'm not entirely against this feature, but its proper implementation seems way more complicated than encoding URLs on the client side. I already see a couple of cases where your implementation will fail:

1. What if both the source and imgproxy URLs have query strings? Of course, this may be treated as an error on the client side, but ideally, this case should be handled by combining the query strings.

2. What if the query string contains percent-encoded characters? imgproxy decodes all the plain URLs, so it will also decode the encoded characters in the query string, making the URL invalid.

> Let's say you have a cookie that only defines Path=/. You have a reverse proxy in front of imgproxy, that reverse proxy will forward the cookies and set forwarding headers, then imgproxy will request a resource from another service - that's where we lose the cookies, "another service" doesn't receive them. Even nginx doesn't strip cookies by default 😄

`Path=/` doesn't look like a problem here. Could you give me a full cookie example, including the `Path` and `Domain` fields, alongside the imgproxy domain? If the `Domain` field contains imgproxy's domain (which is the only case the browser should send that cookie to imgproxy), and your reverse proxy sets the `X-Forwarded-Host` request header, then the current implementation of cookie forwarding should work as expected.

BTW, I made a small fix in the `latest` build that fixes the case where a reverse proxy doesn't set the `X-Forwarded-Host` header.

tonybart1337:
>1.What if both the source and imgproxy URLs have query strings? Of course, this may be treated as an error on the client side, but ideally, this case should be handled by combining the query strings.

>2.What if the query string contains percent-encoded characters? imgproxy decodes all the plain URLs, so it will also decode the encoded characters in the query string, making the URL invalid.

I think you're missing my point here. All I want for imgproxy is to simply preserve the url as it is without any modifications, just pass it to the backend directly without any re-encoding. I don't want to use query parameters of imgproxy in any way.
If I pass url with invalid encoding (with `@` or other unsafe symbols), then it's my fault and not imgproxy

>BTW, I made a small fix in the latest build that fixes the case where a reverse proxy doesn't set the X-Forwarded-Host header.

Thanks. I will try with the latest update

DarthSim:
> I think you're missing my point here. All I want for imgproxy is to simply preserve the url as it is without any modifications, just pass it to the backend directly without any re-encoding.

I do understand your case, but I can't modify imgproxy to cover _only_ your case. imgproxy has thousands of users, and I should keep in mind their cases and the fact that the feature may not work as expected in their cases.

imgproxy always decodes plain URLs as it expects that they may be percent-encoded. Let's assume your imgproxy URL is the following:

```
https://example.com/unsafe/rs:100:100/plain/https://images.example.com/image.jpg?param=the%26value
```

`?param=the%26value` here is a query string of the imgproxy URL, so imgproxy cuts it off as the source URL should be defined in the URL path. So the source URL here is `https://images.example.com/image.jpg`.

Your feature basically makes imgproxy not cut off the query string. In this case, imgproxy's URL parser will include `?param=the%26value` in the source URL. However, the URL parser expects that the URL may be percent-encoded, so it will decode it to `https://images.example.com/image.jpg?param=the&value`, which is an invalid URL. You may say that your query strings will not have encoded characters, yet other users' URLs may.

Another case that came to my mind is where a user may use your feature with Base64-encoded source URLs. In this case, the URL parser will fail to decode the source URL as it will be only partially encoded.

As you can see, there are plenty of corner cases, and not all of them immediately came to my mind. Thus, I don't think covering them all is any simpler than adding a single function call to the client side.

tonybart1337:
I see what you mean now, indeed, there was some code that was still tweakling with the url.
there was also the part that applied the options from the query string

I tested it on your example and it works properly now, thank you

>Another case that came to my mind is where a user may use your feature with Base64-encoded source URLs. In this case, the URL parser will fail to decode the source URL as it will be only partially encoded.

🤔 why would you use this feature with base64, if you use base64 then the url already contains the query string? I would just add a warning to the docs. if you want to pass the url as is without any modifications from imgproxy's side, then use this new feature.
if you want base64, or query string params, or anything else, then don't use this feature

DarthSim (commented):
Sorry, but the last commit has broken more things than fixed.

The correct implementation would be to modify `options.ParsePath` to cut off the query string at the very beginning and then add it to `imageURL` before returning it. If `imageURL` already contains a query string, it should be combined with the added one (simply replace `?` in the beginning with `&`).

Yet again, you're trying to solve a small part of a problem that is usually solved by encoding the source URL; and you do it in a more complex way than just adding a function call to the client side. In all imgproxy-like tools, this problem is solved by encoding the source URL on the client side, because it's more robust. The more I think about this, the less I believe I'd like to see this feature added. There are too many ifs for this feature to be useful.

tonybart1337:
Regarding cookies, here's the setup:
Reverse proxy -> matches /previews -> to imgproxy with base https://images.anotherdomain.com
The cookie is set to `Path=/`.

sample url: 
```
https://example.com/previews/unsafe/rs:100:100/plain/image.jpg?param=the%26value
```
The reverse proxy will passthrough all the headers including the cookies, along with the new forwarding headers.

imgproxy will receive the following url:
```
https://images.anotherdomain.com/unsafe/rs:100:100/plain/image.jpg?param=the%26value
```

`cookieBaseUrl` in this case will be: `https://images.anotherdomain.com:443`, which of course is different from the one that the user sent the request from, hence imgproxy will strip the cookies and not let them through.

I'm sure it's a good security feature to have on by default, but in my case I don't want imgproxy to do more than just a simple preview, therefore I think there should be an option to disable the RFC checks from cookiejar's default implementation and just passthrough the cookies as is.

The reason I'm not passing the entire url is that I don't want to expose the actual sources to the user

DarthSim:
Ah, my bad, I see the problem now. Let me think about it

tonybart1337:
Oh, and BTW, your software currently trusts all the x-forwarded headers that you might throw at it, usually there is always an option to only allow certain CIDRs.

For example: https://doc.traefik.io/traefik/routing/entrypoints/#forwarded-headers

DarthSim:
I added a proper implementation of any cookie passthrough to the `latest` build, see [IMGPROXY_COOKIE_PASSTHROUGH_ALL](https://docs.imgproxy.net/latest/configuration/options#IMGPROXY_COOKIE_PASSTHROUGH_ALL) in the docs. I figured that this would be faster than explaining all the nuances 😅

As for the "query string passthrough" feature, after some consideration, I decided not to include it in imgproxy. The summary of my reasoning for future historians:

1. The problem that is meant to be solved by this feature is traditionally solved by percent encoding on the client side.
2. This feature solves only a part of the problem solved by the percent-encoding of source URLs.
3. I don't want to give users even the slightest idea that they may completely avoid encoding source URLs using this feature.
4. This feature adds new logic to multiple parts of imgproxy while the problem can be solved by a single function call on the client side.
5. This feature is useful only when multiple ifs are met. If, at some point, those ifs stop being met, users may face nasty consequences.

This PR contains a decent implementation of this feature, so anyone can fork it and make their own imgproxy build.

@tonybart1337 Despite the result, I very much appreciate your effort and would love to see further contributions from your side.

PR #1459, opened 2025-07-13
feat: add resizeType contain

Scarbous:
it returns a image in the requested size without croping it. this type acts as the css backround-size type contain

it fills the new whiteSpace with the defined background color


DarthSim:
Hey @Scarbous!

The behavior looks exactly like `resize_type:fit` + `extend:1` (or `extend_aspect_ratio:1` depending on your goal).

Scarbous:
@DarthSim Thank you 🥳

PR #1597, opened 2026-01-06
IMG-81: preserve BPP

gzigzigzeo:


processing/colorspace_test.go
```diff
@@ -0,0 +1,233 @@
...
+			s.Require().NotNil(hash)
+
+			// Match against stored hash
+			s.ImageMatcher.ImageMatches(s.T(), result.OutData.Reader(), "hash", 0)
+		})
+	}
+}
+
+func (s *ColorspaceTestSuite) TestLinearColorspace() {
+	testCases := []colorspaceTestCase{
```
DarthSim: I guess it'd be nice to add a 8-bit sample here to ensure the result is 8-bit.

processing/colorspace_to_processing.go
```diff
@@ -9,27 +10,24 @@ func (p *Processor) colorspaceToProcessing(c *Context) error {
...
 		c.Img.RemoveColourProfile()
 	} else {
 		// vips 8.15+ tends to lose the colour profile during some color conversions.
 		// We need to backup the colour profile before the conversion and restore it later.
 		c.Img.BackupColourProfile()
 
-		if convertToLinear || !c.Img.IsRGB() {
+		// For non-RGB source images, we always need to import ICC profile to get
+		// correct colors.
+		if !c.Img.IsRGB() {
```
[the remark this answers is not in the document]
gzigzigzeo: Probably, it should be done on conversion in scale.go?

vips/vips.c
```diff
@@ -934,23 +966,44 @@ vips_apply_watermark(VipsImage *in, VipsImage *watermark, VipsImage **out, int l
...
+      res = vips_colourspace(in, &t[5], VIPS_INTERPRETATION_RGB16, NULL);
+    }
+
+    if (res) {
+      VIPS_UNREF(base);
+      return 1;
+    }
+
+    in = t[5];
+  }
```
DarthSim: No need to convert anything. We should should just pick a right `compositing_space` and vips will do the conversion under the hood.

vips/vips.c
```diff
@@ -934,23 +966,44 @@ vips_apply_watermark(VipsImage *in, VipsImage *watermark, VipsImage **out, int l
...
+
   if (
       vips_composite2(
-          in, watermark, &t[5], VIPS_BLEND_MODE_OVER,
+          in, watermark, &t[6], VIPS_BLEND_MODE_OVER,
           "x", left, "y", top, "compositing_space", in->Type,
           NULL) ||
-      vips_cast(t[5], &t[6], vips_image_get_format(in), NULL)) {
+      vips_cast(t[6], &t[7], vips_image_get_format(in), NULL)) {
```
DarthSim: `t` has only 7 elements length. We have a leak and probably a segfault here.
We can just revert this when we remove the colorspace conversion

PR #1602, opened 2026-01-15
IMG-92: export logs to monitoring

gzigzigzeo:


monitoring/otel/otel.go
```diff
@@ -140,6 +152,22 @@ func New(config *Config, stats *stats.Stats) (*Otel, error) {
...
+		return nil, err
+	}
+
+	if logExporter != nil {
+		logProcessor := sdklog.NewBatchProcessor(logExporter)
+		o.loggerProvider = sdklog.NewLoggerProvider(
+			sdklog.WithResource(res),
+			sdklog.WithProcessor(logProcessor),
+		)
+		o.logHook = newLogHook(o.loggerProvider, config.LoggerName, logger.Level())
```
DarthSim: I believe loge level should be configurable here like we did for syslog

gzigzigzeo: Could you please elaborate a bit? In what case could this be required?

monitoring/otel/config.go
```diff
@@ -14,19 +14,29 @@ const (
 )
 
 var (
+	protocolMap = map[string]string{
+		"grpc":          "grpc",
+		"http/protobuf": "http/protobuf",
+		"http":          "http",
+		"https":         "https",
```
DarthSim: We could map `http` and `https` to `http/protobuf` here and remove them from protocol checks

PR #1660, opened 2026-05-19
Add `IMGPROXY_MAX_RESULT_WIDTH` config and `max_result_width` processing option

allouis:
## Summary                                                                                                                                                                             
                                                                                                                                                                                         
Adds a new `IMGPROXY_MAX_RESULT_WIDTH` config (and matching `max_result_width` / `mrw` processing option) that caps the result image width independently of its height.
                                                                                                                                                                                         
## Motivation
                                                                                                                                                                                         
`IMGPROXY_MAX_RESULT_DIMENSION` enforces a cap on the *longest* edge of the result image. For tall sources this is undesirable: a 500×5000 image with `IMGPROXY_MAX_RESULT_DIMENSION=2000` is downscaled to 200×2000. The width collapses to 200px, making the image unusable in layouts that need a predictable width.                     
                                                                                                                                                                    
The new option lets operators (or per-request callers) cap width directly while leaving height free to grow with the source's aspect ratio. Both options can be combined, in which case the more restrictive scale factor wins.                                                                                                                                               
                                               
## Behavior                                                                                                                                                                            
                                                               
In `processing/prepare.go::limitScale`, the downscale factor now considers both limits and picks the most restrictive:
                                                                                                                                                                                         
  - If `max_result_dimension` is set and `max(outWidth, outHeight)` exceeds it → `downScale = maxDim / max(w, h)`
  - If `max_result_width` is set and `outWidth` exceeds it → `downScale = min(downScale, maxWidth / outWidth)`                                                                           
  - Aspect ratio is always preserved.                                                                         
                                                                                                                                                                                         
A 500×5000 image with `max_result_width=2000` is left at 500×5000 (already under the cap). A 5000×500 image with the same setting is downscaled to 2000×200.                           
                                                                                                                                                                                         
## Tests                                                                                                                                                                               
                                                                                                                                                                                         
`TestResultWidthLimit` in `processing/processing_test.go` covers:
                                                                                                                                                                                         
  - Width limit above source width → no scaling
  - Width limit caps a wide source                                                                                                                                                       
  - Width limit applies after explicit resize                  
  - Tall (rotated) source is left alone when only height would have tripped the dimension cap — the regression case described above                                                      
  - Both limits set, width limit is the more restrictive                                                                           
  - Both limits set, dimension limit is the more restrictive


DarthSim:
Hey @allouis!

`IMGPROXY_MAX_RESULT_DIMENSION` is a security measure. `IMGPROXY_MAX_RESULT_WIDTH` alone can't serve as a security measure, as it doesn't limit the resulting image height.

allouis:
Hey @DarthSim, thanks for taking a look at the PR!

Over at https://github.com/TryGhost/Ghost/ we're using `imgproxy` for handling resizing all of our images, and we have a requirement to restrict these to a maximum of 2000px wide, but we don't mind about the height - I appreciate that this doesn't serve as a security measure, but it's still a valuable feature for us!

Is there anything else we can do to get this feature added? Perhaps if we added an `IMGPROXY_MAX_RESULT_HEIGHT` option too?

DarthSim:
Why not use `w:2000` then?

allouis:
@DarthSim I'm not sure I completely understand your suggestion - do you mean intercepting the requests and forcing the width to be 2000?

Our URLs are configurable as part of image source sets, so `w` is already varying across breakpoints (e.g. w:400, w:800, etc..) we want to support a variety of image widths, but put an upper bound on the width. So a client could request w:2400 - but we want to cap out at 2000

DarthSim:
I believe there's a piece of code inside Ghost that generates URLs. Checking the width at the time of URL generation, or even at the config parsing stage, seems to be a proper solution here.

If this doesn't work for you, then ok, let's add both MIN_WIDTH and MIN_HEIGHT

allouis:
The problem with that approach is that we leave ourselves open to still having those URLs manipulated.

I think we're going to take a different approach with our fix here though, and instead just limit on the max dimension and resolution.

Thanks for taking your time to look at this!

PR #1667, opened 2026-05-29
feat(): add preset label to request_span_duration_seconds Prometheus …

Ehsanino82:
Add a "preset" label to the request_span_duration_seconds histogram so that span durations can be observed per preset. The applied preset names are extracted from the parsed processing options and passed through the monitoring metadata to the Prometheus StartSpan implementation. When multiple presets are used, the label value is a comma-separated string; when none are used, it is empty.

### DIFF
--- a/handlers/processing/handler.go
+++ b/handlers/processing/handler.go
@@ -2,6 +2,7 @@ package processing
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/imgproxy/imgproxy/v4/auximageprovider"
 	"github.com/imgproxy/imgproxy/v4/clientfeatures"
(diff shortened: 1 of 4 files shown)


Ehsanino82:
@DarthSim Could you take a look at these changes please?

DarthSim:
Hey @Ehsanino82!

I appreciate the effort, but I'd rather not add this. The proposed implementation has several drawbacks:

1. It doesn't set the label for `request_duration_seconds`.
2. It doesn't set the label for `downloading_*` spans. However, presets may include `watermark_url` or `fallback_url` options that affect the downloading.
3. It duplicates presets meta for other meters. They already have this info in `imgproxy.options`.

Fixing this (especially items 1 and 2) would require many changes. On the other hand, presets are a very tiny piece of info about the request. And including info about presets covers quite an edge case, yet still doesn't provide enough info for proper troubleshooting. It would be nice to send the full set of metadata to Prometheus, but it isn't well-suited for that. If you need detailed info about your requests, it's better to adopt request tracing via OpenTelemetry, Datadog, or New Relic.

Still, appreciate your effort 👍

PR #1672, opened 2026-06-04
fix(s3): wrap AssumeRoleProvider in CredentialsCache to prevent per-request STS calls

### LINKED ISSUE #1671: S3 AssumeRole called on every request when IMGPROXY_S3_ASSUME_ROLE_ARN is set, causing STS rate limiting
## Summary

When `IMGPROXY_S3_ASSUME_ROLE_ARN` is configured, imgproxy calls AWS STS `AssumeRole` on every S3 object fetch rather than reusing the token for its lifetime. Under high request volume this exhausts the STS API rate limit (default 15 req/s per account/region), causing S3 fetches to fail with credential errors.

## Root cause

In `storage/s3/storage.go`, the `stscreds.AssumeRoleProvider` is assigned directly to `conf.Credentials` after loading the default config:

```go
creds := stscreds.NewAssumeRoleProvider(sts.NewFromConfig(conf), config.AssumeRoleArn, ...)
conf.Credentials = creds  // no cache
```

`stscreds.AssumeRoleProvider` has no built-in caching. The AWS SDK v2 documentation for `CredentialsProvider` states explicitly:

> A credentials provider implementation can be wrapped with a `CredentialsCache` to cache the credential value retrieved. Without the cache the SDK will attempt to retrieve the credentials for every request.

When credentials are resolved via `awsConfig.LoadDefaultConfig`, the SDK automatically wraps the provider chain in `aws.CredentialsCache`. That wrapping is bypassed here because `conf.Credentials` is replaced after the config is loaded with the raw uncached provider.

## Impact

Every S3 `GetObject` call triggers an `AssumeRole` call. At significant request volume (e.g. high-traffic image serving), this generates hundreds of `AssumeRole` calls per second rather than approximately one per 15 minutes.

## Fix

Wrap the provider in `aws.NewCredentialsCache`:

```go
conf.Credentials = aws.NewCredentialsCache(creds)
```

This caches the STS token and only calls `AssumeRole` when the token is within the expiry window (~10 seconds before expiry), reducing calls from one-per-request to roughly one per 15 minutes.

## Steps to reproduce

1. Configure imgproxy with `IMGPROXY_USE_S3=true` and a valid `IMGPROXY_S3_ASSUME_ROLE_ARN`
2. Generate high request volume (hundreds of concurrent image fetches)
3. Observe `AssumeRole` rate limit errors in AWS CloudTrail or imgproxy error logs

wimnat:
## Context

Closes #1671.

When `IMGPROXY_S3_ASSUME_ROLE_ARN` is configured, imgproxy calls AWS STS `AssumeRole` on every S3 object fetch rather than reusing the token for its lifetime. Under high request volume this exhausts the STS API rate limit (default 15 req/s per account/region), causing S3 fetches to fail with credential errors.

In `storage/s3/storage.go`, the `stscreds.AssumeRoleProvider` was assigned directly to `conf.Credentials` after loading the default config, bypassing the `CredentialsCache` that `awsConfig.LoadDefaultConfig` would otherwise apply automatically. The AWS SDK v2 documentation is explicit:

> A credentials provider implementation can be wrapped with a `CredentialsCache` to cache the credential value retrieved. Without the cache the SDK will attempt to retrieve the credentials for every request.

## Intent

Wrap the `AssumeRoleProvider` in `aws.NewCredentialsCache` so that the STS token is cached for its lifetime (~15 minutes by default) and `AssumeRole` is only called when the token is within the expiry window — reducing calls from one-per-request to approximately one per 15 minutes.

```go
// Before
conf.Credentials = creds

// After
conf.Credentials = aws.NewCredentialsCache(creds)
```

### DIFF
--- a/storage/s3/storage.go
+++ b/storage/s3/storage.go
@@ -67,7 +67,10 @@ func New(config *Config, trans *http.Transport) (*Storage, error) {
 					o.ExternalID = aws.String(config.AssumeRoleExternalID)
 				}
 			})
-		conf.Credentials = creds
+		// Wrap in CredentialsCache so AssumeRole is called once per token
+		// lifetime (~15 min) rather than on every S3 request. Without this,
+		// AssumeRoleProvider has no caching and calls STS on every Retrieve().
+		conf.Credentials = aws.NewCredentialsCache(creds)
 	}
 
 	clientOptions := []func(*s3.Options){


DarthSim:
Hey @wimnat!
Dang, that's a super non-obvious thing. Thanks for catching this!

Could you change a few things before the merge?
1. Change the base to `version/4.0`. We'll release this in the next patch release.
2. Add a comment on why we need `aws.NewCredentialsCache`
3. Remove the test. It basically just checks whether we called a function, which is bad practice for tests; tests should check side effects, not internal logic. The comment from (2) should be enough.

wimnat:
@DarthSim all done 🙂

DarthSim:
Merged. Thanks for the fix! 👍

PR #1684, opened 2026-06-24
Pass through unsupported source types

### LINKED ISSUE #1683: Feature request: Pass through unsupported source types
Currently imgproxy returns 422 for unsupported source types
```
2026-06-24 02:17:16 [WRN] Completed in 14.139263ms /unsafe/quality:80/plain/local:///test.txt request_id=GvIGppo7AbbhRoIA0AXRI method=GET status=422 client_ip=127.0.0.1 error="can't download source image: Source image type not supported"
```

In our case, we serve mixed content (images + arbitrary files like PDFs, TXT, binaries), and we need a unified behavior at the source handling layer rather than per-request routing logic.

**I’m aware of the existing options:**
- [raw processing](https://docs.imgproxy.net/usage/processing#raw) option
- [skip-processing](https://docs.imgproxy.net/usage/processing#skip-processing) option

**However, these do not fully cover this use case:**
- skip-processing applies only to image sources
- raw requires explicitly deciding per-request which URLs must bypass processing (e.g. different handling for images vs documents, quality pipelines, etc.)

## Proposal

What about  new config option? For example `IMGPROXY_PASS_UNSUPPORTED_TYPE=true`

**When enabled:**
- if source type is not supported for processing (non-image / unsupported formats)
- return original file as-is (pass-through)
- do not return 422 error

I can work on a PR for this if the approach is aligned with the project direction

madest92:
Adds `IMGPROXY_PASS_UNSUPPORTED_TYPE` configuration option.

When **enabled**, unsupported source types are returned as-is instead of responding with 422.

Includes tests covering the new behavior.

Resolves #1683

### DIFF
--- a/handlers/processing/config.go
+++ b/handlers/processing/config.go
@@ -13,6 +13,7 @@ var (
 	IMGPROXY_REPORT_IO_ERRORS          = env.Bool("IMGPROXY_REPORT_IO_ERRORS")
 	IMGPROXY_FALLBACK_IMAGE_HTTP_CODE  = env.Int("IMGPROXY_FALLBACK_IMAGE_HTTP_CODE")
 	IMGPROXY_ENABLE_DEBUG_HEADERS      = env.Bool("IMGPROXY_ENABLE_DEBUG_HEADERS")
+	IMGPROXY_PASS_UNSUPPORTED_TYPE     = env.Bool("IMGPROXY_PASS_UNSUPPORTED_TYPE")
 )
 
 // Config represents handler config
@@ -21,6 +22,7 @@ type Config struct {
 	ReportIOErrors          bool // Whether to report IO errors
 	FallbackImageHTTPCode   int  // Fallback image HTTP status code
 	EnableDebugHeaders      bool // Whether to enable debug headers
+	PassUnsupportedType     bool // Whether to pass original image data for unsupported formats
 }
 
 // NewDefaultConfig creates a new configuration with defaults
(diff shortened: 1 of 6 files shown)


DarthSim:
Hey @madest92!

I appreciate the effort, but I can't merge this.

In this implementation, a file of an unsupported type will be downloaded entirely into memory rather than streamed. This leads to multiple consequences:

1. Large files will easily cause an OOM.
2. Passing through large files will be very time-consuming and will lock the worker for a long time while not doing any processing.
3. Video playback won't work, as this implementation doesn't respect the `Range` header.

Also, this implementation doesn't consider the fact that imgproxy Pro generates thumbnails for videos, PDFs, etc.

madest92:
@DarthSim 

Could you please tell me where the error is? I seem to be using the same methods as for images, except for the conversion.

I didn't take the Pro version into account. I couldn't find any reference to PDF in the code.
Supported image types https://github.com/imgproxy/imgproxy/blob/master/vips/vips.go#L143
Other types return 422
```
2026-06-24 14:56:35 [WRN] Completed in 734.447µs /unsafe/quality:80/plain/local:///test.pdf request_id=Qf1WaurvoMYeEqlKHrMwC method=GET status=422 client_ip=127.0.0.1 error="can't download source image: Source image type not supported"
2026-06-24 14:56:46 [WRN] Completed in 565.318µs /unsafe/quality:80/plain/local:///test.mp4 request_id=3ajgpkVmwEixVE5cNZFjV method=GET status=422 client_ip=127.0.0.1 error="can't download source image: Source image type not supported"
```

With new IMGPROXY_PASS_UNSUPPORTED_TYPE=1
```
2026-06-24 15:02:17 [INF] Completed in 10.705342ms /unsafe/quality:80/plain/local:///test.pdf request_id=UOsB_Y0K2wBTvBKyr8g75 method=GET status=200 client_ip=127.0.0.1 image_url="local:///test.pdf" processing_options={ skip_processing=true }
2026-06-24 15:02:19 [INF] Completed in 4.726294ms /unsafe/quality:80/plain/local:///test.mp4 request_id=13t-ISL6IeomWbFa6HOT6 method=GET status=200 client_ip=127.0.0.1 image_url="local:///test.mp4" processing_options={ skip_processing=true }
```

DarthSim:
Closing this along with the original issue. I don't see any real use cases for this.

PR #1699, opened 2026-07-26
fix: gzip response uses the compressed length in async download

ferhatelmas:
When an origin unexpectedly returns gzip content,
wrapGzipBody decodes automatically but preserves
the original content length, which is used to 
limit reads to that size.

Since body is decompressed now:
* if compressed body is smaller, image is silently truncated
* if larger, download reports UnexpectedEOF

After decompression, remove content-length header,
set content length unknown and response uncompressed.

Related to #1497

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [4.0.12] - TBA
+### Fixed
+- Fixed truncated source image downloads when origins return gzip-compressed responses.
+
 ## [4.0.11] - 2026-07-02
 ### Fixed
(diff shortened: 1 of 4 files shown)


gzigzigzeo:
Thank you. LGTM. Just one minor comment: we keep 4.0.x development in the version/4.0 branch. The master branch is now used for the next major release. Could you please rebase your changes on top of version/4.0 when you have a chance?

ferhatelmas:
@gzigzigzeo thanks for your review, updated

gzigzigzeo:
Thank you! Will merge into the upcoming patch release.

Issue #1705, opened 2026-07-29
Non-error panics are logged as "panic: %!w(<nil>)"


ridwanakf:
When a request handler panics with a value that is not an `error` (for example a bare `panic("...")`), the panic recovery in `WithPanic` logs and reports it as:

```
panic: %!w(<nil>)
```

The actual panic value is discarded.

https://github.com/imgproxy/imgproxy/blob/bdeaebcdaff316e00390f6a870008fad8638e5c7/server/middlewares.go#L79-L85

```go
err, ok := rerr.(error)
if !ok {
    err = fmt.Errorf("panic: %w", err)
}
```

On the `!ok` branch `err` is a nil `error`, not the recovered value `rerr`, and `%w` over a nil formats to `%!w(<nil>)`. So the message is lost before it reaches `LogResponse` and the error reporters.

This affects non-error panics such as a bare `panic("...")` from a dependency or cgo wrapper, and imgproxy's own `imageData` refcount guards, which panic with strings.

Suggested fix: format the recovered value, `fmt.Errorf("panic: %v", rerr)` (`%v` because on this branch it is not an `error`). 

I can open a PR with a regression test if you think this worth the changes.

gzigzigzeo:
Thank you for reporting this! Fixed.

ridwanakf:
Hi @gzigzigzeo, Thanks for picking this up, and sorry to reopen the topic. I noticed the merged fix still ends up dropping the value on the non-error branch: there `err` is the nil from the failed `rerr.(error)` assertion, so `fmt.Errorf("panic: %v", err)` logs `panic: <nil>` rather than the recovered value.

Formatting the recovered `rerr` keeps the value, and since `rerr` is `any` it also removes the need for the `//nolint:errorlint`. I opened a small follow-up PR with a regression test that fails on the current `version/4.0` and passes with `rerr`: #1706

If you'd rather leave it as is, feel free to close it. Just want to put it out here

PR #1706, opened 2026-07-30
fix: preserve the recovered value when logging non-error panics

ridwanakf:
Follow-up to the fix for #1705. That change swapped `%w` for `%v`, which removed the `%!w()` artifact, but the recovered value is still dropped on the non-error branch, so these panics now log `panic: <nil>`.

https://github.com/imgproxy/imgproxy/blob/ed78652c4a5d08c4c712c35d45450c95c1b2487a/server/middlewares.go#L79-L85

```go
err, ok := rerr.(error)
if !ok {
    err = fmt.Errorf("panic: %v", err) //nolint:errorlint // err here is not an error, most likely plain string
}
```

On this branch the type assertion failed, so `err` is the nil zero value of `error`; it is `rerr` that holds the recovered value (a plain string, for example). So `fmt.Errorf("panic: %v", err)` formats `<nil>`, and the actual message is still lost before it reaches `LogResponse` and the reporters.

Formatting the recovered value fixes it, and also makes the `//nolint` unnecessary, `rerr` is `any`, so `errorlint` does not flag it:

```go
if !ok {
    err = fmt.Errorf("panic: %v", rerr)
}
```

I added `TestIntoPanicWithNonErrorPreservesValue`: it drives `WithPanic` with `panic("string panic")` and asserts the recovered error carries the value. On the current `version/4.0` it fails with `"panic: <nil>" does not contain "string panic"`; with `rerr` it passes. `go test ./server/...` and `golangci-lint run` pass, and there is a CHANGELOG entry under Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [4.0.13] - TBA
 ## Fixed
 - Fixed inconsistent environment variable names between the code and documentation (`IMGPROXY_AIRBRAKE_ENVIRONMENT`, `IMGPROXY_S3_USE_DECRYPTION_CLIENT`, `IMGPROXY_USE_LAST_MODIFIED` and `IMGPROXY_USE_ETAG`).
+- Fixed recovered non-error panics still being logged and reported as `panic: <nil>` instead of the actual panic value.
 
 ## [4.0.12] - 2026-07-29
 ### Fixed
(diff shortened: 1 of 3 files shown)


ferhatelmas (approved):

ridwanakf:
Hi @gzigzigzeo , a gentle nudge on this when you get a chance.

It is a small follow-up to #1705 : format the recovered rerr rather than the nil err, so non-error panics log their actual message instead of panic: <nil>, with a regression test that fails on the current version/4.0 and passes with the change. I have just rebased it onto the latest version/4.0, so it is up to date and merges cleanly (the changelog entry sits under the existing ## [4.0.13] section).

No rush at all, and no worries if it is not a priority right now; just flagging it in case it slipped off the radar.

gzigzigzeo:
Thanks a lot! Merging.

PR #1714, opened 2026-08-22
chore: drop pkg/errors in favor of stdlib

ferhatelmas:


gzigzigzeo:
Hi there. Unfortunately, we can’t drop the `Cause()` method because it’s used by one of the reporters. We can replace pkg/errors with the standard library; however, this won’t remove the dependency completely, since `pkg/errors` is still used as an indirect dependency somewhere else. Still, it may be worth doing, as the dependency would be dropped automatically once that library stops using it.

ferhatelmas:
> we can’t drop the Cause() method because it’s used by one of the reporters

it seems we're already overriding it [here](https://github.com/imgproxy/imgproxy/blob/88270aef33862951fbe204c51f56e817ce535ed3/errorreport/airbrake/airbrake.go#L40-L44) so it won't be breaking reporters but it's in the public API, that could be a concern to drop. I think probably fine but what do you think? If needed, can revert because it isn't a big deal like the dependency itself.

> this won’t remove the dependency completely, since pkg/errors is still used as an indirect dependency somewhere else. Still, it may be worth doing, as the dependency would be dropped automatically once that library stops using it.

yes, I agree it's still a good direction. Do you want to add a lint rule to prevent it from being used, just in case?

DarthSim:
> it seems we're already overriding it here so it won't be breaking reporters

That's just a single case. Many imgproxy dependencies depend on pkg/errors. `Cause()` is a one-liner, so I don't see a reason for deleting it.

> it isn't a big deal like the dependency itself

The dependency itself is not a huge deal either, to be honest. But I'm ok with removing it from direct deps.

> Do you want to add a lint rule to prevent it from being used, just in case?

I don't think there's a need for that

ferhatelmas:
sounds good, reverted cause drop

ferhatelmas:
@gzigzigzeo shall we merge this?

DarthSim:
Thanks for the contribution!

